PR #25, opened 2015-02-08
A template for making e-books based on the tufte-book latex class and tu...

elbamos:
This is an e-book template based on the tufte-book latex class and tufte-handout rmarkdown template, modified to be more consistent with rmarkdown norms.   

I'm making the request for rticles per a dicussion after I made the request at rmarkdown.  Its seems to fit better with rmarkdown, but its your guys' call of course. 

One of the nice things about this is that the skeleton shows how to do things that can be annoying to figure out, like making a proper bibliography, columns, landscape mode, pretty tables with highlighting, etc.


elbamos:
This is being moved back over from rmarkdown.  I really encourage including some of the information in the skeleton on the rmarkdown help pages, or linking to the pdf -- it took me a long time to figure out how to do things like make a proper bibliography, or print a table where one row is in bold.

Issue #24, opened 2015-02-08
elsarticles class pdf conversion.


yipcma:
It would be great if the elearticle class is supported for the wide-range use in Elsevier journals

PR #27, opened 2015-02-10
Add the Elsevier template (requested in #24)

cboettig:
This adds the Elsevier template files (requested in #24)

My `tex` template includes many options that can be set via top-level yaml in the .Rmd, (e.g. `layout: review`, or `layout: preprint,12pt`, `linenumbers`, `csl` etc),  but perhaps these should be made arguments to the function itself (so that they are set under the output section)?  I'm not sure if that is more or less confusing than setting them at the top level; given that currently a user needs to set most of those variables that are supported by pandoc default templates as top-level metadata.  

Note that this implementation suffers from one weakness: `@` signs in the email will cause Pandoc to wrap the email with an `href` when called from rmarkdown, which uses the non-default option `autolink_bare_uris` and as far as I can tell provides no obvious way to turn that option off.  The use of an `\href` around the email declaration conflicts with the elsarticle.cls definitions of the `\ead()` element.  I'm not clear on the best way to address that and would welcome suggestions.  

Currently I have simply written emails in the template as `bob(at)example.com` to avoid autolinking, but this is a non-ideal solution.

Altering the class file itself is, I think, a worse solution. The `elsarticle.cls` file that already ships with most tex distributions and is found on most major journal platforms (arxiv,  manuscriptcentral, etc) so users who actually submit the `tex` file to the journal would have to be careful to submit the modified class file.  

Perhaps something clever can be hacked into the template itself to redefine the class file's definitions and make this work, but I have no idea how to do so (pending my [question on tex.stackchange](http://tex.stackexchange.com/questions/227393)).

It seems the preferred solution would be simply to turn `autolink_bare_uris` off, but if that's possible in `rmarkdown`, it's beyond me.  https://github.com/rstudio/rmarkdown/issues/372


cboettig:
Looks like all problems with the `@` symbol are now removed, and this template should work as expected.

PR #38, opened 2015-07-10
use UTF-8 as input encoding

krlmlr:
For consistency. I'm not sure if this is really allowed by the affected journals -- if not, we might need to look for another solution, as pandoc will create UTF-8 anyway.

### DIFF
--- a/inst/rmarkdown/templates/acm/resources/template.tex
+++ b/inst/rmarkdown/templates/acm/resources/template.tex
@@ -1,4 +1,5 @@
 \documentclass{acm_proc_article-sp}
+\usepackage[utf8]{inputenc}
 
 \renewcommand{\paragraph}[1]{\vskip 6pt\noindent\textbf{#1 }}
 \usepackage{hyperref}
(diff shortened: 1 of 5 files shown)


jjallaire:
Can we check whether this is in fact allowed by the respective journals before making the change?

krlmlr:
JSS allows inclusion of other packages and doesn't explicitly mention input encoding in its manual. I'm not familiar with the other journals.

We could also insert this as a comment, so that users can easily enable it at their own risk.

jjallaire:
It's up to you, I'll accept either variation.

hadley:
I'm pretty sure most journals will accept utf-8

krlmlr:
It might be easier to ask for forgiveness than to ask for permission in this particular case. The pull request doesn't touch any class or style files.

isezen:
UTF8 encoding seems merged on Jul 11, 2015, but in master branch, I don't see `\usepackage[utf8]{inputenc}` line in acs template. Was it removed in purpose? If I use ACS template and use some Turkish characters in YAML section of the Rmd file, They are not rendered.

PR #57, opened 2016-01-06
Fix errors on Mac

yufree:
@jjallaire Thank you for your comments. I fixed some errors about ACS articles and pass the package check on Mac. However, currently I don't have a machine for windows test. Could you help me about the test on windows?

Thanks,

Miao


jjallaire:
Thanks, but this pull request changes 56 files not just your template! (these are mostly file mode changes). It also regenerates the help with an older version of roxygen. To create the PR I'd suggest you start from current master and do:

```bash
git revert --no-commit 6ecbe466cb6a098eca5edd2a5603515eb00da4d6
```

Then work from there.

yufree:
Thank you very much, @jjallaire 
I think this commit will work.
PS: I know I am not familiar with roxygen and now I find I am not familiar with GitHub, either...

jjallaire:
There are still 66 files changed in the PR!!!! This needs to be much
cleaner for us to take it. We also need you to find a way to test/debug on
Windows as it will likely require iteration and experimentation which will
be extremely unproductive if we are just feeding you error messages. I
recommend that you run a Windows evaluation version within a VirtualBox VM
(that shouldn't cost you anything until the Windows eval times out).

yufree:
Close this dirty PR.

PR #62, opened 2016-03-21
documentclass and classoption of JSS template.

vwmaus:
The commit allows changing documentclass and classoption of JSS template in the Rmd file.


yihui:
Thanks!

Issue #64, opened 2016-03-23
Author addresses in Elsevier


frenchja:
The [Elsevier skeleton](https://github.com/rstudio/rticles/blob/master/inst/rmarkdown/templates/elsevier_article/skeleton/skeleton.Rmd) should probably be updated to include the proper YAML for setting different addresses:

```
address:
  - code: University 1
    address: Department, Street, City, State, Zip
  - code: University 2
    address: Department, Street, City, State, Zip
```

This isn't clear from the RMarkdown provided skeleton unless one digs into the accompanying [skeleton.tex](https://github.com/rstudio/rticles/blob/master/inst/rmarkdown/templates/elsevier_article/resources/template.tex#L172-L174) for the document class.

jjallaire:
Thanks, could you send us a pull request with that change?

frenchja:
Actually, taking a second look, it appears that the `affiliation` should be different from the `code`, which is in line with the Elsevier LaTeX example:

https://github.com/rstudio/rticles/blob/master/inst/rmarkdown/templates/elsevier_article/resources/template.tex#L167-L174

frenchja:
@jjallaire This can be merged.

PR #65, opened 2016-03-23
Add address and footnote examples to Elsevier skeleton.

frenchja:
Updates Elsevier skeleton to add `address` and `footnote` examples.

See https://github.com/rstudio/rticles/issues/64


jjallaire:
Is there another change you wanted to make here or is this good to go?

PR #72, opened 2016-06-10
Add a PNAS template

cboettig:
This provides a template based on the official PNAS LaTeX template (for Overleaf, pnas-new.cls). 

A few notes:

- The LaTeX template uses quite a few custom fields, for the most part these have been turned into yaml header sections (e.g. abstract, author contributions, acknowledgements, etc).  Perhaps there is a way to do this that doesn't clutter the yaml header with so much text. 

- pandoc introduces a dependency on `longtable` when turning markdown tables into LaTeX.  `longtable` is incompatible with any two-column style, including this one.  See https://github.com/jgm/pandoc/issues/1023.  Users can thus not create markdown tables, though can of course use knitr utilities to generate the desired LaTeX table directly anyway, so hopefully this is not an issue.    (It appears that `longtable` is also not always included in other templates here, which could create similar problems for users writing markdown tables.  

- The template download includes several custom `.sty` files.  Additionally I have included some `.sty` files needed here that are not readily found in a tex build that is otherwise sufficient to compile R packages and the default pandoc templates (i.e. the hadleyverse docker image).  I've put these in `skeleton` dir, I'm sometimes unclear what goes in `skeleton` and what goes in `resources`.  Feedback on that or anything else is welcome, will help me write these faster.

- I added a test following the existing pattern.


mikabr:
Was just checking this out and it looks like the author superscripts aren't getting put in correctly -- knitting the default template, every author has a `1` instead of the appropriate letters/numbers.

cboettig:
@mikabr Thanks for catching that. Just pushed a fix for the author footnotes, which should now correctly indicate both affiliations and equal authorship or corresponding author designations.  

Happy for feedback on how the yaml header is structured in the .Rmd file too; I think it's relatively intuitive if a bit verbose, but good to here from someone using fresh without having already looked at the underlying template.  

Also, I'd like the yaml header blocks to be as consistent as possible across journal formats, (i.e., authors are always a yaml list of name, code) even though journals make such different tex templates that may be limited.

paleo13:
Hi, this looks awesome. I was wondering what the progress on this pull request is? Is it likely to change substantially in the future?

cboettig:
I think I'm done fiddling with it now, sorry for the multiple follow-up commits after opening the PR. Of course I'm still open to user suggestions but I think it is now reasonably comparable to the pnas LaTeX template and should be good to go

paleo13:
No worries - I'm planning to use this for a submission. Thanks again for developing this!

PR #96, opened 2016-11-30
Template for the journals from American Meteorological Society

yufree:


inst/rmarkdown/templates/ams_article/resources/template.tex
```diff
@@ -0,0 +1,70 @@
+\documentclass[$col$]{ametsoc}
```
jjallaire: I don't see any of the pandoc includes in this template (e.g. `include-header`, `include-before`, `include-after`). Should these be added?

yufree: Yes, it's convenient! 
I also add `include-header` into the ACS template. 
Thanks!

jjallaire:
There is also `include-before` and `include-after`, see: https://github.com/jgm/pandoc-templates/blob/master/default.latex#L231-L234

Could you also add this to the tests and ensure it passes on all platforms? (see https://github.com/rstudio/rticles/blob/master/tests/testthat/test_formats.R#L42-L48)

tests/testthat/test_formats.R
```diff
@@ -43,10 +43,10 @@ test_format("acm_article")
 test_format("elsevier_article")
 test_format("jss_article")
 test_format("rjournal_article", file_check = FALSE)
-test_format("acs_article", os_skip = "windows")
+test_format("acs_article", file_check = FALSE)
```
yufree: Tests run fine in mac and Linux. 
However, for windows  I found tests failed and I set the `file_check` to false to pass tests both for acs and ams template. Such templates could be used in windows, just do not pass the unit tests. I think the reason would be the path of the figures generated by the templates.
The pnas template also can not pass the test right now and also could be used in windows. I think `os_skip` would work for it while `file_check` does not work.

inst/rmarkdown/templates/acs_article/resources/template.tex
```diff
@@ -63,6 +63,14 @@
 $for(header-includes)$
 $header-includes$
 $endfor$
+$for(include-before)$
```
jjallaire: These headers actually belong within the {document} around the `$body$` (and perhaps other LaTeX that goes at the beginning or end). See https://github.com/jgm/pandoc-templates/blob/master/default.latex#L231-L272

yufree: Thanks a lot and I fixed it.

PR #97, opened 2016-12-01
Template for the IEEE Transaction Journals

Emaasit:
Latex template can be found on the conference site at http://www.ieee.org/conferences_events/conferences/publishing/templates.html


inst/rmarkdown/templates/ieee_article/resources/template.tex
```diff
@@ -0,0 +1,477 @@
...
+$header-includes$
+$endfor$
+% Pandoc toggle for numbering sections (defaults to be off)
+$if(numbersections)$
+$else$
+\setcounter{secnumdepth}{0}
+$endif$
+% Pandoc header
+$for(header-includes)$
+$header-includes$
```
jjallaire: You are including `header-includes` twice.

inst/rmarkdown/templates/ieee_article/resources/template.tex
```diff
@@ -0,0 +1,477 @@
...
+%
+% If IEEEtran.cls has not been installed into the LaTeX system files,
+% manually specify the path to it like:
+% \documentclass[conference]{../sty/IEEEtran}
+
+
+
+
+
+% Some very useful LaTeX packages include:
```
jjallaire: As things stand now there really isn't a way for users to activate any of these classes. You could make these options at the format function level and then pass pandoc arguments along to conditionally enable them.

jjallaire:
Could you also add a call to test the format here: https://github.com/rstudio/rticles/blob/master/tests/testthat/test_formats.R#L42-L48

coatless:
Would it be okay if I contribute or submit the requested revisions to this PR? I would like to get the `IEEEtrans` format available in `rticles`.

Emaasit:
@coatless Yes, please. Go ahead. Due to work overload here at school, I had planned to do the revisions at the end of this month.

coatless:
@Emaasit: Could you give me write access to your fork? Or would you prefer if I PR'd against your fork and on merge those changes would appear here?

Emaasit:
@coatless I have added you as a collaborator to my fork. That gives read/write permissions.

coatless:
Addressed the concerns in 136c182. 

As a side note, @Emaasit, did you want to add in support for the `journal`, `technote`, `peerreview`, and `peerreviewca` paper modes? Or just keep it focused on the `conference` mode?

Emaasit:
@coatless Thanks for those patches. Two more fixes are needed before it's perfect.
- [ The sections should be number: How about we just add `numbered_sections: true` in the yaml header] 
- [The References should be numbered ] 

Regarding support for other articles formats, I think we should add only the `journal` format for now.

coatless:
Good catches by the way.

1. Why not remove this variable entirely as IEEE requires section to be numbered? 
2. Regarding citations:
- The IEEE style will not work with `biblatex` since the package is anti `.bst`. However, it seems as if using [`\usepackage[style=ieee]{biblatex}`](http://ctan.mackichan.com/macros/latex/contrib/biblatex-contrib/biblatex-ieee/biblatex-ieee.pdf) to get around this hiccup is not possible either. I'll test this out on a different machine tomorrow. An alternative solution to address this issue is to just fall back and use the `bibtex` backend. 

- Meanwhile, the `natbib` setup can easily be handled by:

```
\usepackage[numbers]{natbib}
\bibliographystyle{IEEEtranN}
```

where `IEEEtranN` is the [`IEEEtranN.bst` from `IEEEtran`](https://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/bibtex).

Regarding supporting the `journal` in addition to `conference` mode, I'll add a few variables in that allow a paper mode to be specified. The main task to supporting different modes is being able to handle the different styles of author listings and allowing for IEEE keywords.

espinielli:
This PR looks quite a mature contribution...is there any plans to merge it?

PR #109, opened 2017-04-10
Plos template

BlackEdder:
This pull request adds a template for plos (computational biology). It is based on http://journals.plos.org/ploscompbiol/s/latex and the elsevier template.


BlackEdder:
The only problem I had is that in the plos latex example line numbering is not enabled for the references, but I could not figure out how that can be done in rmarkdown/pandoc

jjallaire:
Is the file PLOS-submission-eps-converted-to.pdf supposed to be part of the skeleton?

jjallaire:
Did you run the tests on Mac, Windows, and Linux? (I've noticed that sometimes formats fail on one or another platform due to vagaries of LaTeX).

BlackEdder:
The PLOS-submission-eps-converted-to.pdf file seems to be created from the original eps file. I thought I hadn't commited it, will delete it.

Afraid I only have easy access to linux machine, so only tested there.

jjallaire:
Just tested on Mac and Windows, all looks good!

Issue #107, opened 2017-03-09
Royal Society of Statistics


carlganz:
http://www.rss.org.uk/RSS/Publications/Journals/Journals_get_involved/RSS/Publications/Journals_sub/Get_Involved.aspx

Would you consider a pull request for RSS publications?

yihui:
#110 has been merged. Thanks!

PR #110, opened 2017-04-12
Royal Society of Statistics

carlganz:
Based on #107


jjallaire:
Could you add this to the tests here (https://github.com/rstudio/rticles/blob/master/tests/testthat/test_formats.R#L42) and run the tests on any platforms you have access to a working LaTeX installation on? (I like to check on Linux, Windows, and OS X if possible)

PR #111, opened 2017-04-17
ASA journal style

hadley:
This looks good from my end. What else do you need for rticles?


jjallaire:
Could you add a test here (https://github.com/rstudio/rticles/blob/master/tests/testthat/test_formats.R#L42) once this is ready to go. I usually like to run these tests on all platforms before merging b/c the distinct LaTeX environments can easily create incompatibilities.

hadley:
I refactored `test_format()` — it didn't work with `asa_article` because `output_file` was a full path, rather than a file name. I couldn't quite figure out what you were trying to do, but I think I've captured the intent and simplified the code a little (and all the tests pass for me)

hadley:
(I also alphabetised the test formats so it's more obvious where you should add your test)

jjallaire:
Looks good, thanks!

PR #113, opened 2017-05-04
Added a template for Statistics in Medicine

ellessenne:
I added a template based on the LaTeX class of Statistics in Medicine.
Some tables are broken, if produced by `kable()`; nevertheless, `xtable()` works fine (example in `skeleton.Rmd`).
I tried running the tests, they failed for a bunch of templates (I suspect because of MiKTeX on a managed university laptop); knitting the file works fine, though.


jjallaire:
Test pass for me. Thanks!

PR #114, opened 2017-05-09
Updated the SiM template

ellessenne:
Hi,
I updated the template for Statistics in Medicine for a greater adherence to its LaTeX class. 
The template knits fine on my side, as before some tests do not succeed.
Thanks!

### DIFF
--- a/R/sim_article.R
+++ b/R/sim_article.R
@@ -19,5 +19,7 @@
 #' @export
 sim_article <- function(...) {
   inherit_pdf_document(...,
-                       template = find_resource("sim_article", "template.tex"))
+                       template = find_resource("sim_article", "template.tex"),
+                       highlight = NULL,
+                       citation_package = "natbib")
 }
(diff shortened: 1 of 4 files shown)


ellessenne:
I am puzzled by the `mathtime` package requirement: it renders on MiKTeX (tested on Windows 7), not on TeXLive (tested on OS X 10.11.6) - or is it just me?

jjallaire:
The tests are actually also failing for me on OS X with this error:

```
output file: Untitled.knit.md

! LaTeX Error: File `mathtime.sty' not found.

! Emergency stop.
<read *> 
```

I think we need to remove the mathtime.sty requirement (or somehow bundle it) since we want our templates to work out of the box on an OS X LaTeX installation. Could you submit another PR with this change?

ellessenne:
Thanks @jjallaire, I will do that asap!

jjallaire:
I am hoping to submit to CRAN soon but am waiting on your change. If you can't make it over the next couple of days I'll just unexport this format for the time being. Let me know.

ellessenne:
I will have access to a Mac for testing on Monday, would that be too late?

jjallaire:
Yup, that would be fine!

PR #118, opened 2017-05-20
Adds a preamble variable to asa template

coatless:
Allows for custom LaTeX declarations similar to rjournal.


coatless:
Thanks for the quick merge!

PR #119, opened 2017-05-30
Tables in SiM template

ellessenne:
Hi!
I added support for custom tables packages to the Statistics in Medicine template, via adding the `header-includes` option in the YAML header.
Best wishes,

Alessandro

### DIFF
--- a/inst/rmarkdown/templates/sim_article/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/sim_article/skeleton/skeleton.Rmd
@@ -15,6 +15,7 @@ year: 2017
 abstract: "This paper describes the use of the \LaTeX `simauth.cls` class file for setting papers for Statistics in Medicine using Rmarkdown."
 acknowledgements: "We thank [Stack Overflow](https://stackoverflow.com/) for the invaluable support to our research."
 keywords: Class file; \LaTeX; Statist. Med.; Rmarkdown;
+tablesenv: ctable
 bibliography: bibfile.bib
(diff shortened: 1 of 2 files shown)


inst/rmarkdown/templates/sim_article/skeleton/skeleton.Rmd
```diff
@@ -15,8 +15,12 @@ year: 2017
 abstract: "This paper describes the use of the \LaTeX `simauth.cls` class file for setting papers for Statistics in Medicine using Rmarkdown."
 acknowledgements: "We thank [Stack Overflow](https://stackoverflow.com/) for the invaluable support to our research."
 keywords: Class file; \LaTeX; Statist. Med.; Rmarkdown;
+header-includes: 
+- \usepackage{ctable}
```
jjallaire: I don't think this should be part of the default skeleton (i.e. if we want it always included it should be in the template proper).

ellessenne: I see; I put it in the skeleton to provide the user with an example, I always find it easier to remove parts I do not want than digging up the right syntax for parts I may want to add at some point. Makes sense? If not, I will remove it and make another pull request.
Thanks @jjallaire!

jjallaire:
Rather than automatically inserting that, I would just provide the example
YAML code along with the code below that demonstrates it's use.

ellessenne:
Ok, I moved the example to the text and removed it from the YAML header!

PR #124, opened 2017-06-11
elsarticle formatting

krivit:
This patch removes some of the hard-coded style customizations in the template, restoring settings such as the positioning of floats, the margins, and paragraph indentation to their class defaults.

Also, it adds the `$preamble$` directive, so that the user could customize those settings if they like.


jjallaire:
@cboettig You created the original template here, could you review and approve?

cboettig:
:+1:

Yup, this looks good to me. Thanks @krivit

PR #127, opened 2017-06-27
PeerJ format

zkamvar:
Hello,

This is my submission for the [PeerJ format](https://peerj.com/about/author-instructions/). It currently has an issue with the placement of the "References" header, but other than that, everything should be cromulent. 

I modified the template from Overleaf: https://www.overleaf.com/latex/templates/latex-template-for-peerj-journal-and-pre-print-submissions/ptdwfrqxqzbn

I've also included pandoc-style cross-references (e.g. `\@ref(tab:table1)`), which requires the users to use `bookdown::pdf_document2()`, which I've documented in the help and the skeleton. Otherwise, they can switch back to using `\ref{}` with the standard latex tables and figures.


jjallaire:
Looks good!

jjallaire:
Could you open a new PR to add this to the package tests? If convenient it would also be great if you could run the test on Windows, OS X, and Linux and confirm it passes.

zkamvar:
Hi! I just realized that I had muted this repo at some point in the past and didn't see this until now. 
The version you pulled in should have the tests: https://github.com/rstudio/rticles/pull/127/files#diff-4f8540d05c0675b39d119ee57b3511d4

I have tested this on OSX and am in the process of testing it on Windows

jjallaire:
Okay, great. Let me know if tests pass on Windows then I'll merge.

zkamvar:
It appears to work on windows as well.

jjallaire:
Great, thanks!

yihui:
@zkamvar Please note that I just removed the `base_format` argument, since I noticed that people started doing this in other output formats following this format (e.g., #208 #172), which is really not necessary. To make use of **bookdown**'s features, now you need to use:

```yaml
output:
  bookdown::pdf_book:
    base_format: rticles::peerj_article
```

This way, we don't have to patch every single format in this package to support **bookdown**'s features.

Thanks!

Issue #120, opened 2017-05-30
Guidance on including *.sty files, assumptions about tex environment, and thoughts on travis?


cboettig:
@jjallaire 

Hi JJ,

Just wanted to get your take on how contributors should be handling dependencies wrt `.sty` files.  

It would be great if the README or package SysReqs gave some more guidelines on what the tex environment should be (i.e. ideally something less comprehensive than texlive-full, which weighs in at >2 GB unpacked). 

Previously I've dropped .sty files that I didn't have in a 'vanilla' tex environment into the template's [skeleton dir](https://github.com/rstudio/rticles/tree/master/inst/rmarkdown/templates/pnas_article/skeleton), though this clutters the user's working directory a bit when they adopt the template.  It would be nice for those sending PRs to know what they should assume visa vi the tex environment, so they can add only the .sty files that are necessary.  (or maybe the assumption is always full texlive and I should never be adding sty files to a PR?)  

Relatedly,  I'd like to make sure our [rocker/verse](https://github.com/rocker-org/rocker-versioned/blob/master/verse/Dockerfile#L27-L30) image included the necessary style files to compile any template in `rticles`.  I can do this somewhat manually just by seeing what tests fail on the image of course.  

Lastly, it would be nice to have some automated tests on the repo here, and having a bit more definition on what the tex environment should be would help in setting up the .travis.yml, if you're open to a PR adding a travis file.  That would also set a nice default benchmark for when a contributor should be providing additional .sty files.  

Thanks for your thoughts on this!

jjallaire:
I think the problem of Tex for R / R Markdown is that the users aren't
actively maintaining their Tex libraries but rather using TeX as a black
box dependency. If we want to teach users how to do TeX package management
we can, but I think given the depth and diversity of templates the sanest
thing to do is to just install the full TeX environment.

Another alternative would be for TeX based R Markdown formats to declare
their TeX dependencies, and then have some functions available to install
them in the user TeX library (or even to auto-install them as MikTeX does).
I sadly don't have time to take on something that ambitious right now!

dougmet:
I think this issue covers my question too. I'm starting work on a PR to add [APS journals](https://journals.aps.org/) widely used by the physics community. All of their LaTeX is covered by the [revtex](https://www.ctan.org/pkg/revtex4-1?lang=en) package on CTAN.

[...]

PR #130, opened 2017-07-06
Add travis

zkamvar:
I've added a .travis.yml file. It currently fails because of a few cls files that are either mis-named ore are not in the right place:

 - [acs_article](https://travis-ci.org/zkamvar/rticles/jobs/250523297#L2349)
 - [ams_article](https://travis-ci.org/zkamvar/rticles/jobs/250523297#L2413)
 - [pnas_article](https://travis-ci.org/zkamvar/rticles/jobs/250523297#L2480)

It's also failing due to some templates using either xtable or lubridate.

Nevertheless, I'm creating this PR for feedback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+# R for travis: see documentation at https://docs.travis-ci.com/user/languages/r
+
+language: r
+r:
+  - release
+  - devel
+
+sudo: false
+cran: http://cran.at.r-project.org
+
+cache: packages
+
+before_install:
+  - Rscript -e 'update.packages(ask = FALSE)'
+
+notifications:
+  email:
+    on_success: change
+    on_failure: always
+
+env:
+  global:
+    - NOT_CRAN: true
+    - R_CHECK_ARGS="--as-cran --timings"
+    - R_CHECK_TIME="TRUE"
+    - R_CHECK_TESTS="TRUE"
+    - _R_CHECK_TIMINGS_="0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ cache: packages
 
 before_install:
   - Rscript -e 'update.packages(ask = FALSE)'
+  - tlmgr install endfloat
 
 notifications:
   email:
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,2 +1,3 @@
 ^.*\.Rproj$
 ^\.Rproj\.user$
+^\.travis\.yml$
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
 
+[![Travis-CI Build Status](https://travis-ci.org/rstudio/rticles.svg?branch=master)](https://travis-ci.org/rstudio/rticles)
+
 ### Installation
 
 You can install and use **rticles** from CRAN as follows:


yufree:
For `acs_article` and `ams_article`, you need to install extra CTAN packages. You could add the following code to `.travis.yml` to pass the tests:

```
before_install:
    - tlmgr install achemso ametsoc
```

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+cache: packages
+
+before_install:
+  - Rscript -e 'update.packages(ask = FALSE)'
+  - tlmgr install endfloat achemso ametsoc
+
+notifications:
+  email:
+    on_success: change
+    on_failure: always
```
yihui: I think these are the defaults, so probably not needed.

.travis.yml
```diff
@@ -0,0 +1,28 @@
+# R for travis: see documentation at https://docs.travis-ci.com/user/languages/r
+
+language: r
+r:
+  - release
+  - devel
+
+sudo: false
+cran: http://cran.at.r-project.org
```
yihui: Could you use `https://cran.rstudio.com`?

zkamvar:
Since some of the formats use lubridate and xtable, should I also add these to the `Suggests:` in the DESCRIPTION?

yihui:
I think lubridate can be removed. The code can be simply replaced by `format(Sys.Date(), '%Y')` and `format(Sys.Date(), '%m')`.

I don't have an opinion on xtable. Any of these methods is fine to me:

- adding it to Suggests
- install it on Travis only (`before_install`)
- `if (!require("xtable"))` install.packages("xtable") in Rmd

zkamvar:
Okay! I'll try to get to these within the next couple of days (but the branch is open for edits if you need it before then 😺 ).

DESCRIPTION
```diff
@@ -29,4 +29,4 @@ License: GPL-3
 Imports: utils, rmarkdown, knitr, yaml
 SystemRequirements: GNU make
 RoxygenNote: 6.0.1
-Suggests: testthat
+Suggests: testthat xtable
```
yihui: missing comma here

zkamvar: Rupaul screaming like I'm screaming

zkamvar:
I'm done fiddling with this for tonight. Does anyone have a way to resolve all of these dependency issues all at once? The current LaTeX error I'm getting is:

```
LaTeX Error: File `needspace.sty' not found.
```

cboettig:
I'm probably missing something, but why not just include texlive-full?  I know it's big and slow, but apt installs are all cached on .travis, right?

yufree:
Hi, I have a dirty solution by searching `\usepackage` in this repo like [this](https://github.com/rstudio/rticles/search?p=4&q=%5Cusepackage&type=&utf8=%E2%9C%93). Then grab all packages by regular expression and add them to `before_install`. However, some packages might be installed by default.

yufree:
I steal some codes from [here](https://github.com/ghseeli/grad-school-notes/blob/d1d3945e9bfc8bcdd3c08919717aad835897524d/.travis.yml) if you want to install the texlive-full.

```
before_install:
- |
  if [[ ! -e ./texlive/bin/x86_64-linux/tlmgr ]]; then
    wget http://mirrors.ctan.org/systems/texlive/Source/install-tl-unx.tar.gz -O - | tar -x --gzip
    mv $(ls | grep install) install-tl
    ./install-tl/install-tl -profile texlive.profile
  fi
- export PATH=$(pwd)/texlive/bin/x86_64-linux:$PATH
- sudo apt-get install texlive-full
- sudo apt-get install latexmk
- tlmgr install $(cat .tlmgr-dep)
- tlmgr update --self
- tlmgr update --all
```

zkamvar:
I'm trying out @cboettig's suggestion because it seems to be the simplest answer so far. If I were worried about things that were big and slow, I would have never gone into academia 😆 . I'm about to get on a 🛫 , so I hope this works 🤞.

zkamvar:
🔥🔥LaTeX errors🔥🔥

yihui:
@cboettig Apt dependencies are probably not cached, and Travis does not recommend caching Debian packages: https://docs.travis-ci.com/user/caching/#Things-not-to-cache

I have been annoyed by this TeXLive issue for several years, and I'm relatively sure that I can come up with a more efficient solution, i.e. you do not need a 5G texlive-full yet you can use most packages. After I finish the blogdown book, I'll try to take a stab at it.

cboettig:
@yihui thanks. I feel like @jimhester gave me some pointers on some clever LaTeX configurations in r-travis a while back but now I cannot find them.  

I'm also happy to see alternatives to texlive-full, fwiw `rticles` passes check on our `rocker/verse` image ([Dockerfile](https://github.com/rocker-org/rocker-versioned/blob/master/verse/Dockerfile#L26-L49)) which adds about a [300 MB layer](https://microbadger.com/images/rocker/verse), though I'm not sure if all the rticles formats have complete unit tests.  

I asked about the assumed `tex` environment back in #120 with an eye to adding `.travis` to the repo and documenting the package's tex dependencies for users (e.g. me), but JJ's advice there was that it was better to just assume a full tex environment.

yihui:
LaTeX support for R on Travis was based on my early work at https://github.com/yihui/ubuntu-bin and I think Jim improved my build scripts and hosted the package elsewhere (I stored it via Github releases).

I wanted to say exactly what I said here back when I saw your discussion in #120. Installing texlive-full is indeed the easiest solution at the price of wasting a lot of disk space. It is just whether you care about the disk space (and download time in the case of Travis). I guess most people probably don't care, but I'm not very comfortable with the fact that a large portion of the texlive-full package is mostly useless to most users (in particular, the docs). Installing texlive-full is just like installing the full CRAN to solve the problem of missing R packages. I believe we can come up with a ~100Mb "base" texlive, and give a simple instruction on how to install missing packages like `install.packages()` in R. Debian has made this awfully difficult and awkward. It should have been `tlmgr install`. That is pretty much the only thing you need to remember about TeXLive package management. No `sudo`, or whatsoever.

cboettig:
Thanks @yihui , I'm with you 100% on avoiding the bloat, clearly don't want the our docker images to be much larger than they need to be (which is the main reason why we separated the tex components out of `tidyverse` and into the separate `verse`.  `tlmgr` is awesome, but I suspect @jjallaire is right that the average R user will find it much more difficult to manage the tex dependencies of `rticles` with `tlmgr` than they do managing R dependencies with `install.packages()`, particularly since knowing what templates need what style files can be rather opaque behind the layers of Rmd, pandoc etc.

yihui:
The worst case is to emulate MikTeX through an R package, i.e. automatically install missing LaTeX packages (using a more awkward approach, though, by parsing the error log). I believe we can solve 99% of the problems with a significantly smaller texlive package plus an R wrapper package. I have done some work in `rmarkdown:::latexmk`. So the basic pieces have already been there. I just need some time to put them all together.

yihui:
@cboettig I spent a week working on this problem and here is what I have got so far: https://github.com/yihui/tinytex Please help me test it. Note that you need a special branch of rmarkdown at the moment:

```r
devtools::install_github(c("yihui/tinytex", "rstudio/rmarkdown@feature/tinytex"))
```

Two major advantages of TinyTeX:

1. It is small.

2. If the LaTeX distribution is TeX Live (e.g., TinyTeX), missing LaTeX packages should be automatically installed by default.

PR #134, opened 2017-08-31
Update peerj article citation format

zkamvar:
I have updated the PeerJ article citation format to be more pandoc-tic.

### DIFF
--- a/R/peerj_article.R
+++ b/R/peerj_article.R
@@ -26,7 +26,7 @@
 #' }
 #'
 #' @export
-peerj_article <- function(..., keep_tex = TRUE, citation_package = "natbib",
+peerj_article <- function(..., keep_tex = TRUE, citation_package = "none",
                           base_format = rmarkdown::pdf_document) {
   if (inherits(base_format, "character")){
     FMT <- eval(parse(text = base_format))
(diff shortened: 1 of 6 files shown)


jjallaire:
@yihui Does this look okay to you?

yihui:
@zkamvar What is the requirement of citations in PeerJ? Does it disallow natbib or biblatex? I don't think using pandoc-citeproc is a good idea for LaTeX/PDF output.

zkamvar:
@yihui There are [no explicit citation requirements for PeerJ](https://peerj.com/about/author-instructions/). While this change sets pandoc-citeproc as the default, it does not exclude natbib or biblatex as possibilities (due to the additions to [template.tex](https://github.com/rstudio/rticles/pull/134/files#diff-67a4d3359ec7587bf447c8f7b9d839f9)). I can change the `citation_package` argument default back to "natbib" if you would like.

The previous iteration of `peerj_article()` [hard-coded natbib in the cls file](https://github.com/rstudio/rticles/pull/134/files#diff-0017d0466549c61e517440d6823b2468L45), which caused problems when attempting to use other options. 

For reference here is the citation with

 - pandoc-citeproc:
   
image

 - natbib

image

 - biblatex

image

zkamvar:
Any update on this?

zkamvar:
ping

PR #144, opened 2017-12-06
solve potential font issue at https://tex.stackexchange.com/questions…

yufree:
…/267675/pdftex-error-pdflatex-file-ecbx0800-font-ecbx0800-at-600-not-found

Only asa and jss templates have this issue.


yufree:
I checked the file and found this template use `LF` or `$`. I have no idea why git show so much diff by adding one line.

PR #148, opened 2018-01-07
prevent extra spaces between Chinese and non-Chinese characters in code blocks

heavenzone:
prevent extra spaces between Chinese and non-Chinese characters in code blocks
the same issue like here https://github.com/yihui/bookdown-chinese/commit/c70ef766b117f3
and here https://stackoverflow.com/q/47990152/559676

### DIFF
--- a/inst/rmarkdown/templates/ctex/resources/default.latex
+++ b/inst/rmarkdown/templates/ctex/resources/default.latex
@@ -94,12 +94,12 @@ $if(lhs)$
 $endif$
 $if(highlighting-macros)$
 $highlighting-macros$
+\RecustomVerbatimEnvironment{Highlighting}{Verbatim}{commandchars=\\\{\},formatcom=\xeCJKVerbAddon}
 $endif$
 $if(verbatim-in-note)$
 \usepackage{fancyvrb}
 \VerbatimFootnotes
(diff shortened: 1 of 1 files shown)


jjallaire:
@yihui Does this look okay to you?

yihui:
I briefly looked at it yesterday and I felt the added line should be inserted elsewhere (conditional on syntax highlighting, i.e., when syntax highlighting is not enabled, this line should not be added). I'll take another look later today.

heavenzone:
@yihui you're right. i've tested some conditions.
if the rmd file has no 

    ``` 
    <code> 
    ```


,  it would cause an error.
so it sould be put into `(highlighting-macros)`. like:
```
$if(highlighting-macros)$
$highlighting-macros$
\RecustomVerbatimEnvironment{Highlighting}{Verbatim}{commandchars=\\\{\},formatcom=\xeCJKVerbAddon}
$endif$
```

yihui:
@heavenzone You are absolutely correct. Please revise your PR and test it. Thanks!

https://github.com/heavenzone/rticles/edit/patch-1/inst/rmarkdown/templates/ctex/resources/default.latex?pr=/rstudio/rticles/pull/148

Issue #66, opened 2016-03-23
Specify journal in Elsevier YAML


frenchja:
Feature Request:  We could allow the author to specify which journal they've submitted the preprint to in the RMarkdown YAML, instead of hard coding it to Elsevier: 

https://github.com/rstudio/rticles/blob/90286b14141b957b68b3db6b5d0a5eafefb1e4ce/inst/rmarkdown/templates/elsevier_article/skeleton/elsarticle.cls#L398

jjallaire:
If you are motivated to get these changes in I suggest either a PR or contact the original author of the Elsevier format (@cboettig) to see if they are willing to make the enhancements (we don't actively maintain formats that are contributed).

cboettig:
@frenchja Thanks for the suggestion.  Unfortunately as you see, Elsevier chose not to make that a configurable option in their `tex` template -- that's hard-coded in the `cls` file by Elsevier.  You should feel free to modify that class file in your own use, but rmarkdown/pandoc don't provide a way to customize class files with yaml (since they aren't meant to be customized in that way anyhow).  There is probably some clever override you could add to your tex header (which rmarkdown does support).

@jjallaire feel free to close this since I don't think it's immediately actionable.

yihui:
Fixed via #152.

PR #152, opened 2018-01-29
fix journal footer line in Elsevier template

cboettig:
Here's a quick fix to allow users to set the journal name in the footer line, as requested in #66.  Apologies for not realizing how to do this earlier!  Also, let me know if I should delete the extraneous `skeleton.tex` and `skeleton.pdf` file in the template's `skeleton` directory, as these are basically just wasting space.  Maybe some the `numcompress.sty`  file should also be dropped.  (as tinytex has basically removed the need to try and include sty files in rticles templates...)


cboettig:
deleted `skeleton.pdf` and `skeleton.tex`.  No hits for `tinytex::tlmgr_search("/numcompress.sty")` so I left that file in.

Issue #132, opened 2017-08-10
Printing issue for JSS style


benkeser:
I'm running into a situation where printing certain objects does not render to .tex properly. I can't sort out what is causing the error and I'm not certain whether this is an issue with `rticles` or with pandoc. Here's a MWE:

RMD document:

---
author:
  - name: A girl has no name
    affiliation: An affiliation
    address: >
      An address
    email: \email{anemail@aninstitution.edu}
    url: http://website.com
title:
  formatted: '\pkg{apackage}'
  plain:     'apackage'
  short:     'apacakge'
abstract: >
  blah blah blah
keywords:
  formatted: [key1]
  plain:     [key2]
documentclass: jss
classoption: article
output: rticles::jss_article
---

# Introduction
```{r, echo = FALSE, message = FALSE}
if(!("SuperLearner" %in% row.names(installed.packages()))){
  install.packages("SuperLearner")
}
library(SuperLearner)
```
This non-exported function prints fine.

```{r}
SuperLearner:::.SL.require
```
This one does not. 

```{r}
SuperLearner:::predict.SL.glm
```

When I build, the relevant part of the .tex file looks like

```latex
This one does not.

\textbackslash{}begin\{CodeChunk\}

\begin{CodeInput}
R> SuperLearner:::predict.SL.glm
\end{CodeInput}

\textbackslash{}begin\{CodeOutput\} function (object, newdata, \ldots{})
\{ if (is.matrix(newdata)) \{ newdata = as.data.frame(newdata) \} pred
\textless{}- predict(object = object\$object, newdata = newdata, type =
``response'') pred \} \textless{}environment:
namespace:SuperLearner\textgreater{} \textbackslash{}end\{CodeOutput\}
\textbackslash{}end\{CodeChunk\}
```

END RMD DOCUMENT

Any guesses as to why the `\textbackslash{}` is being introduced?

yihui:
Yes, it is caused by the dollar sign. It is a known issue (see #10 and #49), but no one seems to have time to fix it. Technical background: https://github.com/jgm/pandoc/issues/2493

JanMarvin:
Could we add a latex raw attribute to hook_chunks like: ` ```{=latex} ` if not using `kable`? Added that in jss_article, but it breaks if a chunk uses `cache=TRUE`. This should fix the problem above and new a behavior of pandoc 2.0 which treats underscores similar to the dollar sign:

```
# test.Rmd
\begin{foo}
a_b
\end{foo}`
[...]

PR #161, opened 2018-04-02
added ^ $ _ character problem fix

jaredhuling:
proposed fix to #132 

It's a bit hackish, especially removing the \end{verbatim} line, but it seems to work


yihui:
Sorry, I closed it by accident. I'll work on the fix for #132 soon. Thanks!

Issue #90, opened 2016-10-13
Honour header-includes


hughjonesd:
It would be nice if these formats would honour the header-includes metadata in .Rmd files. This could be done by adding

```
$for(header-includes)$                                                                                                 
$header-includes$                   
$endfor$
```

to the relevant templates.

lioumens:
@hughjonesd Thanks! I thought I was going insane when my header-include statements were not working.

jjallaire:
@yihui Would it be straightforward for us to add these or are there side effects we are worried about? (seems like if no one uses the the in_header as it stands we wouldn't break anyone by doing this).

@hughjonesd I believe this makes sense. Sometimes template authors don't think to add this but they certainly should. We'd certainly welcome a PR with these additions though!

yihui:
I believe it should be safe to add the support to the existing templates in this package.

jjallaire:
Okay, @hughjonesd could you give us a PR for this?

hughjonesd:
Can do.

David

PR #171, opened 2018-05-15
Add header-includes to various TeX templates

hughjonesd:
See #90. This passes tests on my machine, though I haven't added tests to confirm header-includes are working. I get some warnings about "input string XXX is invalid in this locale".


yihui:
Thanks!

Issue #125, opened 2017-06-20
Feature Request: Support in Rticles for Using Additional Journals, such as Biometrics


skywalker26:
I've done some academic writing with the Rticles package and I love it! I'm trying to write a journal article to publish in Biometrics, but am having a difficult time figuring out how I would go about adding in the Biometrics document class and such. Is this something that you'd be open to developing further?

dlebauer:
(I think this issue could be renamed 'how can I contribute a template for a new journal?)

It would be great to have a step-by-step guide to adding a new template. Perhaps this could be added to the package documentation, the README or CONTRIBUTING.md

In addition to the instructions, it could help to have a generic article type - this would serve two purposes: 1) a directory named ``inst/rmarkdown/templates/generic_article` that someone wishing to contribute could copy to `inst/rmarkdown/templates/journalname_article` and 2) for folks who don't want to contribute a template or doesn't know the target submission journal could get started writing a first draft.

Based on a few pull requests, It seems it _should_ be straightforward to add a new template, but perhaps not https://github.com/rstudio/rticles/pull/83 https://github.com/rstudio/rticles/pull/147, PeerJ https://github.com/rstudio/rticles/pull/127

_main content_

* add `journalname_article()` function to R/journalname_article.R 
  * what are requirements here? is this mostly copy-paste and find-replace?
* add inst/rmarkdown/templates/journalname_article/resources/template.tex
  * how to convert a journal provided latex template to a Rmd template? Seems lots of use of `$` to define variables
* add inst/rmarkdown/templates/journalname_article/skeleton/skeleton.Rmd
* add inst/rmarkdown/templates/journalname_article/template.yaml
* which sty / bst / cls / csl etc files to add?   
   * #83 adds natbib.sty - isn't that usually available by default?


_housekeeping_

* Update namespace (could be done by devtools::document()
* update NEWS
* Update README w/ link to supported journal 
* Add tests (what are we testing for?)
   * example: https://github.com/dleutnant/rticles/blob/80fa67ca34f58270dbe2de622d331023e0becd64/tests/testthat/test_formats.R
* gitignore?

daltonhance:
I was able to create Biometrics template in my fork. I used this package of pandoc templates as a starting point: https://github.com/sachsmc/pandoc-journal-templates That package has other statistics journals that wouldn't be too hard to translate to rticles. My fork has all the main content @dlebauer listed, but I would need a little help walking through the housekeeping for a pull request.

[...]

PR #175, opened 2018-06-21
Add template for the MNRAS (Monthly Notices of the Royal Astronomical Society) article.

oleskiewicz:
This creates a template for a submission to MNRAS journal, based on their LaTeX template file.  I've added most of the housekeeping as outlined [here](https://github.com/rstudio/rticles/issues/125#issuecomment-359955260), but wasn't sure what more is there to be done.


R/mnras_article.R
```diff
@@ -0,0 +1,29 @@
...
+mnras_article <- function(...,
+                        keep_tex = TRUE,
+                        md_extensions = c(),
+                        fig_caption = TRUE){
+  pdf_document_format(...,
+                      keep_tex = keep_tex,
+                      md_extensions = md_extensions,
+                      format = "mnras_article",
+                      template = "template.tex",
+                      csl = "mnras.csl")
```
jjallaire: You need to use `system.file()` to refer to these (as the current code assumes they are in the current working directory)

oleskiewicz:
Thanks @jjallaire, I'm just starting learning R so this is all a bit new to me!

I have changed the `template` and `csl` references using the `find_resource` helper function, just as `plos_article` [does it](https://github.com/rstudio/rticles/blob/371fee0e383daeb1b40d5914dfbdf434430dcb6a/R/plos_article.R#L24), I hope this is better now?

jjallaire:
You should also run the tests locally (you can do this with `devtools::test()`). When I run your test I get this failure:

```r
unused argument (csl = "/Users/jjallaire/packages/rticles/inst/rmarkdown/templates/mnras_article/resources/mnras.csl")
1: rmarkdown::render(testdoc) at /Users/jjallaire/packages/rticles/tests/testthat/test_formats.R:30
2: create_output_format(output_format$name, output_format$options)
3: do.call(output_format_func, options)
```

I think you want to call `pdf_document_format()` rather than `pdf_document()`

oleskiewicz:
Oh I see -- it seems like the LaTeX environment is trying to pick up the example figure (`example.png`), but it assumes the `.tex` extension and fails due to some relative path.

I'll change the way figure is rendered in the skeleton (it's my understanding that including figures through `knitr` is preferred anyway?), hopefully this will fix it.

Sorry it keeps failing, it built successfully for me.

oleskiewicz:
The build fails for the same reason as previously:

    Package pdftex.def Error: File `example.png' not found: using draft setting.

I have checked other templates, and for instance `amq_article` passes the Travis build while including figures from external files in skeleton.  Similarly, there are no issues with the template locally.  How do I include the `example.png` in the test render?

inst/rmarkdown/templates/mnras_article/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,120 @@
...
+worry about the exact layout, which will be handled by the publishers.
+
+Figures are referred to as e.g. Fig. \ref{fig:example_figure}, and tables as
+e.g. Table \ref{tab:example_table}.
+
+\begin{figure}
+  % To include a figure from a file named example.*
+  % Allowable file formats are eps or ps if compiling using latex
+  % or pdf, png, jpg if compiling using pdflatex
+  \includegraphics[width=\columnwidth]{example.png}
```
yihui: You included `example.png` here but the file is not included in the PR. You should either include a real file, or comment out this line.

oleskiewicz:
The tests have passed!  Thank you for all your help -- let me know if there is anything else that needs fixing.

PR #183, opened 2018-08-15
Update template.tex

tormodb:
Added option to turn coloring of hyperlinks off. Passing values to colorlinks or urlcolor from yaml does work, but I found that only setting colorlinks=false actually rendered the links (urls as well as refernces to tables and figures) in black color. I have not done this before, so I am not sure if this will do the trick...

### DIFF
--- a/inst/rmarkdown/templates/elsevier_article/resources/template.tex
+++ b/inst/rmarkdown/templates/elsevier_article/resources/template.tex
@@ -105,7 +105,7 @@ $endif$
             bookmarks=true,
             pdfauthor={$author-meta$},
             pdftitle={$title-meta$},
-            colorlinks=true,
+            colorlinks=$if(colorlinks)$$colorlinks$$else$true$endif$,
             urlcolor=$if(urlcolor)$$urlcolor$$else$blue$endif$,
             linkcolor=$if(linkcolor)$$linkcolor$$else$magenta$endif$,
             pdfborder={0 0 0}}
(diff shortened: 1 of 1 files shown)


inst/rmarkdown/templates/elsevier_article/resources/template.tex
```diff
@@ -105,7 +105,7 @@
             bookmarks=true,
             pdfauthor={$author-meta$},
             pdftitle={$title-meta$},
-            colorlinks=true,
+            colorlinks=$if(colorlinks)$$colorlinks$$else$true$endif$,
```
yihui: I'm confused: if `colorlinks=false`, it seems that it will turn on coloring? How about

```
$if(colorlinks)$colorlinks=true,$endif$
```

?

I know this will change the default behavior of this template in previous versions (default will be "off" instead of "on"), but if users want to turn on coloring, they can set `colorlinks: true` in YAML.

tormodb: Hi, Yihui. I may have not understood how to code this properly. What I wanted to achieve was to avoid the hardcoded ```colorlink=true```in the template, and be able to pass ```true```or ```false```from yaml. For my own use (and perhaps most others) it would almost never make any sense to color links in a paper for submission to a journal, so I would be fine with the default corresponding to the links being rendered black (both urls and crossreferences).

yihui: Okay, I think we are in agreement. Let's change this line to `colorlinks=$if(colorlinks)$true$else$false$endif$,`, which means the default will be black.

yihui:
Could you test it?

```r
devtools::install_github('rstudio/rticles')
```

tormodb:
Brilliant, now works as expected; Default behavior is now to "color" links black, if I write colorlinks: true in yaml, they go back to blue as they did originally. 

Thanks for the superquick work on this!

yihui:
Glad to know! You are welcome!

PR #184, opened 2018-08-22
IEEETrans authors layout enhancements

espinielli:
I am writing an article where some authors have no `department` which would result in a blank line:

with-blank-line

This little  proposal make it possible to avoid the (ugly) blank line:

no-blank-line

### DIFF
--- a/inst/rmarkdown/templates/ieee_article/resources/template.tex
+++ b/inst/rmarkdown/templates/ieee_article/resources/template.tex
@@ -335,9 +335,7 @@ $if(natbib)$
 \bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
 $endif$
 $if(biblatex)$
-\usepackage{biblatex}
-\bibliographystyle{IEEEtran}
-\bibliography{IEEEabrv}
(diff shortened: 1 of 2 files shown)


espinielli:
The latest addition is about supporting 3 authors layouts:

- author-columnar, the one in the original template according to IEEETran for max 3 authors
  author-columnar

- institution-columnar, where authors are grouped per institution (see `Tyrell Inc.`)
  institution-columnar

- wide, as described in IEEETrans template
  ieeetrans-wide

espinielli:
`author-columnar` is left as per original implementation in terms of authors data in YAML not to disrupt previously used versions of the package.
BUT in terms of code it would be beneficial to have a single YAMS structure for affilations as per `institution-columnar` and `wide` and just modify the way it is traversed for the 3 different layouts.

I will push a further commit in this direction. It can eventually dismissed if backward compatibility is deemed more important than reduce implementation complexity.

espinielli:
Unified YAML affiliation for the 3 author layout styles.

espinielli:
BibLaTeX was not working, the last commit fixed it for me (well, Stackoverflow helped a lot there ;-)

yihui:
Thanks!

PR #189, opened 2018-09-21
Promote new, preferred DOI resolver

katrinleinweber:
The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Yes, a bit ironic that they would change the URL of their service, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org). Because the old links with the `dx` subdomain continue to work, there is no urgent need to do anything.

However, I'd hereby like to suggest to follow the new recommendation and update the demo  link.

PS: Since this is similar to a typo fix, I didn't consider the PR template checklist applicable.

### DIFF
--- a/inst/rmarkdown/templates/acm_article/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/acm_article/skeleton/skeleton.Rmd
@@ -51,7 +51,7 @@ references:
     given: Martin
   container-title: Nature Materials
   volume: 11
-  URL: 'http://dx.doi.org/10.1038/nmat3283'
+  URL: 'https://doi.org/10.1038/nmat3283'
   DOI: 10.1038/nmat3283
   issue: 4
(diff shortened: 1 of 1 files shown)


katrinleinweber:
PS: What about the [out-dated DOI-generating codes in all the `.cls` files](https://github.com/rstudio/rticles/search?l=TeX&q=dx.doi)? Would you patch them here, if I pushed the commits into this PR, or do they absolutely, positively have to be patched upstream?

katrinleinweber:
You're welcome :-) Whenever they apply these updates upstream, there shouldn't be any conflicts, at least. I notified at least the JSS folks about this already, but since the other .cls' contribution processes seem to be e-mail- or SVN-based or unspecified, I'm not sure where to start.

Maybe summarily emailing them a link to the diff, so they can follow-up as well?

PR #190, opened 2018-09-21
Point new DOI hyperlinks to preferred resolver

katrinleinweber:
@yihui: Thanks for merging #189. Here are the other suggested updates.


katrinleinweber:
You're welcome, and thanks for merging :-)

PR #208, opened 2019-03-12
asa_article() gains a new argument 'base_format'

earowang:
I would like to have `bookdown::pdf_document2()` for the ASA article template, where I could be able to use `\@ref()` to generate a nice crossed-reference html document too.


earowang:
Oh, wasn't aware of that. Thanks.

PR #213, opened 2019-03-15
ENH: Documentclass for frontiers_article

muschellij2:
I've added this `documentclass` argument to the options for `frontiers_article`, so you can do:
```
output: 
  rticles::frontiers_article:
    documentclass: frontiersHLTH
```
So that you can switch the `documentclass` if necessary.  I had to change the test slightly.


muschellij2:
Thanks - any reason a top-level?  I figured it'd be better as a argument of the function because then you can actually test it: https://github.com/rstudio/rticles/pull/213/files#diff-f334dae09aebe2e5ecc7a62a11791e43R50

PR #222, opened 2019-04-19
edited elsevier_article to use with natbib

JohannesFriedrich:
The option `citation_package: natbib` in the elsevier_article was not possible because the template initialized two bibliographystyles. In addition to that the parameter `bibliography` was not recognized because it was names `biblio-files` in the template.
Furthermore I removed the biblatex commands, because in the elsevier_article class no biblatex support is given and will always give an error.
The only restriction is to name the bibliography file *without* the extension *.bib due to the construction of natbib.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,9 @@
 rticles 0.8
 ---------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


JohannesFriedrich:
Done!

PR #229, opened 2019-06-02
Add JOSS and JOSE format

noamross:
While JOSS and JOSE take markdown submissions, this seemed the best place to put the pipeline so that R authors could locally preview the rendered PDF.

@karthik @arfon @eveskew

---

- [x] Unless you have done it in any other RStudio's projects before, please sign the [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). You can send the signed copy to <jj@rstudio.com>.

_Previously signed_

- [x] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [x] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname_article/resources/template.tex`.

- [x] Add a skeleton article `inst/rmarkdown/templates/journalname_article/skeleton/skeleton.Rmd`.

- [x] Add a description of the template `inst/rmarkdown/templates/journalname_article/template.yaml`.

- [ ] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. Please keep the number of new files absolutely minimal, and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include one hundred items in it without using all of them).

- [x] Update NEWS.

- [x] Update README with a link to the newly supported journal.

- [x] Add a test to `tests/testit/test-formats.R`.

- [x] Add your name to the list of authors `Authors@R` in DESCRIPTION. You don't need to bump the package version in DESCRIPTION.

Thank you!


R/joss_article.R
```diff
@@ -0,0 +1,91 @@
...
+
+  rmarkdown::pandoc_available('2.2', TRUE)
+
+  logo_path <- find_resource("joss_article",
+                             paste0(journal, "-logo.png"))
+  journalname <- ifelse(journal == "JOSS",
+                        "Journal of Open Source Software",
+                        "Journal of Open Source Education")
+
+  template_variables <- c(
```
yihui: How about setting these variables in the YAML frontmatter of the Rmd document? That way, you won't have to worry about escaping certain special characters in the values of these variables (although I guess it would be unlikely that these values could contain characters like double quotes). It will also make the R function much simpler.

R/joss_article.R
```diff
@@ -0,0 +1,91 @@
...
+    pandoc_args = pandoc_args,
+    citation_package = "none"
+  )
+
+  base$knitr$opts_chunk$dev <- c("png")
+  base$knitr$opts_chunk$dpi <- 300
+  base$knitr$opts_chunk$fig.width <- 6
+  base$knitr$opts_chunk$fig.height <- 4.15
+  base$keep_md <- keep_md
+  base$clean_supporting <- !keep_md
```
yihui: All above options can be passed to `pdf_document_format()` so you don't need to hack the `base` object.

R/joss_article.R
```diff
@@ -0,0 +1,91 @@
...
+      paste0("page=", page),
+      "graphics=true"
+    )
+
+  template_variables <- c(rbind(rep("-V", length(template_variables)),
+                                template_variables))
+
+  pandoc_args <- c(
+    template_variables,
+    "--csl", find_resource("joss_article", "apa.csl")
```
yihui: I hope the `csl` option could be set in the YAML metadata of the Rmd document, too.

R/joss_article.R
noamross: Will do, except `keep_md`, which is not supported in `pdf_document()` (https://github.com/rstudio/rmarkdown/issues/1001)

R/joss_article.R
noamross: I wanted to avoid doing this because I wanted the format to compile correctly on a standard `paper.md` without modifying the YAML.

R/joss_article.R
noamross: I wanted to avoid needing these variables in the YAML so that a standard `paper.md` would compile without setting them, but now I realized that the document will compile without most of them.  The logo and journal name still make sense to define here, but I can drop the rest.

noamross:
OK, optimized the images, and simplified the function now that I realize that the article will compile without all those variable set.  I have left the `csl` coded into the function.  I think that the most common workflow will be for authors with already-created `paper.md` files to want to compile them to check references, so want to keep the need to change the YAML minimal.

noamross:
Just checking in. Is this version mergable?

yihui:
Sorry, I haven't had a chance to look at the revision yet.

R/joss_article.R
yihui: All other templates in this package include `csl` under the `skeleton/` directory (instead of `resources/`) so the csl file will be copied over when creating a new article using `rticles::draft()`, e.g. https://github.com/rstudio/rticles/tree/master/inst/rmarkdown/templates/acm_article/skeleton Then in Rmd: https://github.com/rstudio/rticles/blob/master/inst/rmarkdown/templates/acm_article/skeleton/skeleton.Rmd#L15

I'm aware of the potential disadvantages of this approach (e.g. the `csl` file may change in the upstream), but I'd like this template to be consistent with other templates. Similarly, I hope `logo_path`, `journal_name`, and `graphics` variables could all be set in the YAML frontmatter of `skeleton.Rmd` (the png logos need to be moved to `skeleton/` accordingly).

R/joss_article.R
yihui: I just added support for `keep_md` in **rmarkdown**, so you can get rid of this hack, too.

R/joss_article.R
```diff
@@ -0,0 +1,64 @@
...
+  base$clean_supporting <- !keep_md
+
+  base$post_knit <- function(metadata, input_file, runtime, ...) {
+    if (!is.null(metadata$authors)) {
+      citation_author <- ifelse(
+        length(metadata$authors) > 1,
+        paste(metadata$authors[[1]]$name, "et. al."),
+        metadata$authors[[1]]$name
+      )
+      return(c("-V", paste0("citation_author=", citation_author)))
```
yihui: Again, I wish this variable could be set in YAML. I'm definitely being selfish here: I want the R source code of this package to be as simple as possible, and I guess the price for users to pay is not that high (typing out a name).

noamross:
It's not the upstream that I'm concerned with here as the _downstream_.  With JOSS, the `.md` file will be submitted, not the LaTeX or PDF, and then the file goes through their automated build system.  So I'm concerned that user settings in the YAML will clash with that.  Perhaps @arfon can comment here - if users set additional fields such as `csl:` in their YAML, will that break the JOSS/JOSE build in any way?

karthik:
Arfon will have to weigh in but yes setting csl in the yaml will not be used in the automated build system. It's hard coded [here](https://github.com/openjournals/whedon/blob/a1723d160bb6e92deadf606e193b9adc4dadf153/lib/whedon/processor.rb#L98)

yihui:
Since `csl` is hard-coded in the build system, user's setting in YAML will be ignored. See https://pandoc.org/MANUAL.html#citation-rendering:

> Set the `csl` field in the document’s metadata to `FILE`, overriding any value set in the metadata.

arfon:
> Arfon will have to weigh in but yes setting csl in the yaml will not be used in the automated build system. It's hard coded here

Yep, we don't support custom CSL definitions for JOSS papers.

noamross:
I've moved `csl` and other variables to the YAML rather than the function.  I have kept the logic in the function to switch between JOSE and JOSS articles.  This is slightly more complex than the simplest article functions but it's DRYer than having two formats for JOSS and JOSE.

Issue #217, opened 2019-04-08
Is elsarticle-num supported for bibliographystyle?


Freguglia:
I'm writing an article for a paper which requires references to be cited with numbers and square brackets `[1]`. I remember using `\bibliographystyle{elsarticle-num}` with the file `elsevier-num.bst` worked when using pure latex, but I can't figure out how to achieve this with `rticles`. Is it possible? Simply adding `\bibliographystyle{elsarticle-num}` doesn't actually change the bibliography style. I also trying adding it to the yaml header but nothing seems to change it.

nuest:
@Freguglia I took a look at your problem. If you don't mind, please try if the branch `elsevier_cls` in [my fork](https://github.com/nuest/rticles/tree/elsevier_cls) solves the problem for you (tell me if you don't know how to install that one), when you configure 

```
csl: elsevier-without-titles.csl
```

in the YAML header of your Rmd file.

@yihui I moved the `.csl` files into the skeleton directory, comparable to other journals that use CSL, and removed the `\bibliographystlye` and `\biboptions` from the template, see [changes](https://github.com/rstudio/rticles/compare/master...nuest:elsevier_cls?expand=1). Do you think this is feasible?

yihui:
@nuest I have no idea about the journal requirements, so please feel free to make the decision by yourself. If you submit a PR, I'll just merge it. Thank you!

nuest:
I suggest to wait for feedback from @Freguglia. 

@JohannesFriedrich as a recent contributor to this template might also have an article at hand to check this change out.

Note to self: #222 also related to this (did some changes also done in `elsevier_cls`, merged into `elsevier_cls` branch).

Freguglia:
@nuest Sorry for the late response. Your fork solves the problem, the references are cited with numbers and square brackets as needed for the journal after adding `csl: elsevier-without-titles.csl` to the YAML header.

PR #233, opened 2019-06-19
Use csl for Elsevier

nuest:
See #217


nuest:
With `csl: elsevier-harvard.csl`:

image


With `csl: elsevier-without-titles.csl`:

image

PR #240, opened 2019-08-20
add unambiguous flag for Copernicus template detection

nuest:
After the first authors started using the template, I discussed "detecting" submissions based on rticles on the side of the publisher. They would be interested to learn how many authors use the template and have added the flag added by this PR to their publication system to be automatically detected.

No rush, but would be great if this could be included in the next release.


yihui:
I have submitted the new version to CRAN. If everything goes well, it should be on CRAN tomorrow.

PR #255, opened 2019-10-31
Added shared footnotes to elsevier_article()

salauer:
Updated the template for elsevier_article to allow for two authors to share the same footnote (e.g. "these authors made equal contributions") and updated the corresponding skeleton to demonstrate how to use a shared footnote. To do this, I mimicked the style of shared affiliations. (This is my first pull request to an RStudio repo. Any feedback you have for me would be much appreciated!)

### DIFF
--- a/inst/rmarkdown/templates/elsevier_article/resources/template.tex
+++ b/inst/rmarkdown/templates/elsevier_article/resources/template.tex
@@ -147,12 +147,14 @@ $preamble$
 
   \title{$title$}
   $for(author)$
-  \author[$author.affiliation$]{$author.name$$if(author.footnote)$\corref{c1}$endif$}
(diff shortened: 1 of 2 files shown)


salauer:
Good point. I had started with that, but then wasn't sure if there would be some confusion between author.footnote and footnote.code (since author.affiliation links to address.code). I just checked locally and that doesn't appear to be a problem. Do you want to me to submit a new pull request?

salauer:
I think this is ready to be merged.

wulms:
What is about multiple footnotes for one author? E.g. the corresponding author contributed equally with another author?

salauer:
@wulms That's a good observation. I had assumed it worked the same way as with affiliations, but it clearly doesn't. Should make an issue for this.

PR #269, opened 2020-01-20
Set better default values for references on IEEE template

swhaat:
- IEEE template works by default with biblatex, it makes sense to use it as a default on the Rmd file so that users can benefit directly from its usage.
- As the package biblatex is used, the options style=numeric and sorting=none are unnecessary, as these parameters are the default for references on the IEEEtran.bst which is shipped with the ieee template., and its used when creating the references style.
- A consequence of biblatex is the citations have to be used in native latex form \cite{...}, as pandoc conversion will create \autocite or \textcite..
- biblio-title is removed as the references title can be defined directly on the Rmd file, because the references will be appended at the end of the file.

### DIFF
--- a/inst/rmarkdown/templates/ieee_article/resources/template.tex
+++ b/inst/rmarkdown/templates/ieee_article/resources/template.tex
@@ -335,7 +335,7 @@ $if(natbib)$
 \bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
 $endif$
 $if(biblatex)$
-\usepackage[backend=bibtex,citestyle=ieee,style=numeric$if(citation_sorting)$,sorting=$citation_sorting$$endif$]{biblatex}
+\usepackage[backend=bibtex,citestyle=ieee]{biblatex}
 $if(bibliography)$
 $for(bibliography)$
 \addbibresource{$bibliography$}
@@ -632,7 +632,7 @@ $endif$
 $endif$
 $endif$
 $if(biblatex)$
-\printbibliography$if(biblio-title)$[title=$biblio-title$]$endif$
+\printbibliography[heading=none]
 
 $endif$
 $for(include-after)$
@@ -640,5 +640,4 @@ $include-after$
 
 $endfor$
 \end{document}
-
 
--- a/inst/rmarkdown/templates/ieee_article/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/ieee_article/skeleton/skeleton.Rmd
@@ -66,6 +66,7 @@ abstract: |
   The abstract goes here.
   On multiple lines eventually.
 
+bibtex: true
 bibliography: mybibfile.bib
 output: rticles::ieee_article
 #citation_sorting: none   ## used as sorting option of the biblatex package (if selected)
@@ -192,12 +193,12 @@ The conclusion goes here.
 Acknowledgment {#acknowledgment}
 ==============
 
-The authors would like to thank...
+The authors \cite{Feynman1963118} would like to thank...
 
 Bibliography styles
 ===================
 
-Here are two sample references: @Feynman1963118 [@Dirac1953888].
+Here are two sample references: \cite{Feynman1963118,Dirac1953888}.
 
 \newpage
 References {#references .numbered}
--- a/inst/rmarkdown/templates/ieee_article/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/ieee_article/skeleton/skeleton.Rmd
@@ -69,7 +69,6 @@ abstract: |
 bibtex: true
 bibliography: mybibfile.bib
 output: rticles::ieee_article
-#citation_sorting: none   ## used as sorting option of the biblatex package (if selected)
 ---
 
 Introduction


inst/rmarkdown/templates/ieee_article/skeleton/skeleton.Rmd
```diff
@@ -66,9 +66,9 @@ abstract: |
   The abstract goes here.
   On multiple lines eventually.
 
+bibtex: true
```
yihui: What does this option do?

inst/rmarkdown/templates/ieee_article/resources/template.tex
```diff
@@ -335,7 +335,7 @@
 \bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
 $endif$
 $if(biblatex)$
```
yihui: Does IEEE require using `biblatex`? If so, I feel the function `rticles::ieee_article` should have a default option `citation_package = "biblatex"`.

swhaat: IEEE officially uses bibtex [IEEE template](https://www.ieee.org/conferences/publishing/templates.html), and since biblatex its the natural successor it makes sense to use biblatex as a default

inst/rmarkdown/templates/ieee_article/skeleton/skeleton.Rmd
swhaat: sorry, typo, should be `biblatex`... i inserted it because i wanted that the template uses biblatex and the `citation_package = "biblatex"` does not produce the correct references..
might be worth having another look..

swhaat:
I do want that the IEEE template uses biblatex as a default, but taking a look a it now, i think the `biblatex` entry on the Rmd seems a bit hacky... I will investigate further why the `citation_package = "biblatex"` does not behave correctly and will amend this request to a more standard solution

swhaat:
As the idea is to use biblatex as a default I modified the `template.tex` so that if the user gives a bibliography file, it then uses biblatex with that file as a default.

DunLug:
Hello everyone,
This PR seems ok to me, and keeping the IEEE template as simple as possible is a good idea for scientifics with basic computer science knowledge.
I wonder if the natbib option use would work correctly, or have some side effects with these modifications. Maybe the natbib usage could be removed if biblatex is good enough for IEEE papers.
Finally, I would suggest to customize the markdown reference conversions ([@citation] or @citation) to produce \cite commands. I don't know if it is possible but it could be nice to keep rmarkdown document as clean as possible.

swhaat:
I have generally met people that use biblatex for their papers, that is why i suggested that we make it a standard.. i do confess not to try natbib, so im unaware of the results...
Regarding the @citation usage, as i said pandoc does not convert @citation to \cite{}, but rather to \textcite{} or \autocite{}, which might be problematic when using xelatex or in some cases pdflatex.. so i opted to use the direct version of \cite{} which doesnt need pandoc translation and it would just work as one might expect... 
I also support to keep the rmakrdown as clean as possible, but in this case, or unless someone has another workaround i would keep the \cite{}, so that the template works out-of-the-box, and people can directly start to write their content...

DunLug:
I made some research about a way to convert pandoc citation syntax to \cite{} commands. It seems that it is not possible easily so it's better to stay with raw latex \cite commands for now. A quick-and-dirty trick could be to renew \autocite and \textcite commands in template to have \cite behaviour but I don't know if this could create unwanted effects.

About natbib / biblatex, it seems that there're two options :
natbib defined with biblio-style, biblio-files, biblio-title options
bibliography defined (list of bibliographic ressources) that trigger biblatex
If we consider keeping natbib, maybe these modifications could improve clarity :
- use of a single option for the citation engine with natbib or biblatex values
- use of a single field for bibliography / biblio-files
- use biblio-style and biblio-title for both engine if needed, or remove them

DunLug:
I made some quick tests about my previous post. I added :
`
rmarkdown::render('input_file.Rmd', rticles::ieee_article('final', pandoc_args=c('--biblatex')))
`
to disable citeproc filter and add --biblatex option. This should be possible using yaml header parameter citation_package but I didn't find how.

Then, this generates a tex file with `[@test]` and `@test` \autocite and \textcite commands. To consider them as \cite command I've added renewcommands in the skeleton file:
`
\renewcommand{\autocite}{\cite}
\renewcommand{\textcite}{\cite}

\begin{document}
`

The issues with this method are:
- The --biblatex option is given at hand to the function call. It could be interesting to make all processing pipeline using one citation_package
- I don't know how to make multiple citations as with \cite{test1,test2} using markdown syntax
- The bibtex bbl file isn't generated so users needs to create pdf file using pdflatex, bibtex and pdflatex twice

swhaat:
to add bibllatex to the file just:
```yaml
output:
  rticles::ieee_article:
    citation_package: biblatex
```
the renew command does seem pretty hacky, but the compatibility might be an issue.. if pandoc translates the @citation to something other than \textcite or \autocite.. then more renew commands will have to be added.. seems to require too attention to what pandoc might do..
I think the raw \cite{} serves a better global solution even thought it doesnt look as polished as plain markdown, but it will probably create less compatibility issues

DunLug:
About the renew command, I consider that forcing pandoc to convert [@cite] to \cite would be great, while rticles users are not necesseraly aware of the need to use \cite command.
For now, it's effectively not needed to bother ourselves with this issue.

I've checked pandoc documentation again and I think that there is an elegant way to choose between natbib and biblatex behaviour, using pandoc options:
```
pandoc -t latex --bibliography=biblio1.bib --bibliography=biblio2.bib --biblatex -V biblio-title:Titre -V biblio-style:numbered -V biblatexoptions:sorting=none file.md
```
This allows to simplify the template with all work made on pandoc side. But this would probably need another pull request.

@yihui After discussion I'm OK with this merge 👍

umarcor:
FTR:

- https://tex.stackexchange.com/questions/201582/natbib-ieeetran-and-alphanumeric-citation-is-there-an-ieeetransan-style-availa/244651#244651
- https://tex.stackexchange.com/questions/202963/how-to-cite-author-in-ieee-format/202964#202964

Hence, to make the transition easier, I would suggest to set `\usepackage[backend=bibtex,style=ieee-alphabetic,natbib=true]{biblatex}` by default.

Regarding bibtex vs biber as backend, in [this](https://github.com/LarsAsplund/github-facts) bookdown document/project, [biblatex](https://github.com/LarsAsplund/github-facts/blob/main/01-index.Rmd#L42) and [`ieee.cls`](https://github.com/LarsAsplund/github-facts/blob/main/01-index.Rmd#L8) are defined. That installs `biber`: https://github.com/LarsAsplund/github-facts/runs/962397200?check_suite_focus=true#step:3:1247. However, I don't know whether it is actually used.

umarcor:
> * biblio-title is removed as the references title can be defined directly on the Rmd file, because the references will be appended at the end of the file.

@swhaat, this conflicts with bookdown-demo, where the 'References' title is shown for HTML output only: https://github.com/rstudio/bookdown-demo/blob/master/06-references.Rmd. Moreover, depending on the output, the word might change between 'References' (article) and 'Bibliography' (report or book); hence, users would need to work around it.

cderv:
@swhaat What is the status on this PR ? do you plan to work on it again or is it ready to merge  ?

swhaat:
As we had discussed with @DunLug it was about how pandoc translates the `@ref` into `\textcite` or `\autocite`.. and this is behavior could lead to errors as some latex processors do not like those `\XXXXcite` commands. And if the user just uses direct `\cite` this is no longer a problem because pandoc just process it as it is..

so, @cderv after the agreement and discussion with @DunLug this PR was/is ready.

inst/rmarkdown/templates/ieee_article/resources/template.tex
```diff
@@ -334,14 +334,14 @@
 \usepackage{natbib}
 \bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
 $endif$
```
cderv, line 336: @swhaat @DunLug as it seems `biblatex` should be use by default, should this be removed ? 
It seems to me it can cause issue with the new default if someone provide `natbib` option.

cderv:
I see that the template is using the `\cite` package. Currently it is note by default. 
https://github.com/rstudio/rticles/blob/623fb4668d798620a05da87495b74f4ecf501d11/inst/rmarkdown/templates/ieee_article/resources/template.tex#L95-L114

If we activate `biblatex` by default, then if I set `with_cite: true` for this format. I'll get an error. 
```tex
! Package biblatex Error: Incompatible package 'cite'
```

(We got the same type of error if we activate `natbib` - see https://github.com/rstudio/rticles/pull/269#discussion_r472012525)

For citation, this package is mentioned to be used with [`IEEEtrans`](https://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/) : 
http://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/IEEEtran_HOWTO.pdf
and because `IEEEtrans` contains  

I am mainly new to this as I discover this issue but I am wondering: why use natbib or biblatex, and even make biblatex the default here ? It does not seem to be part of the template.

inst/rmarkdown/templates/ieee_article/resources/template.tex
DunLug: It seems that it could be conflict in that case. If we want to allow the natbib option, we could replace these lines with an else statement :
$if(natbib)$
\usepackage{natbib}
$else$
\usepackage[backend=bibtex,citestyle=ieee]{biblatex}

$for(bibliography)$
\addbibressource{$bibliography$}

In that case, maybe another conflict could occur on lines 634-635. As the if statement should only occur with biblatex package, we could replace these with : 
$if(natbib)$
$else$
$if(bibliography)$
\printbibliography[heading=none]
$endif$
$endif$
$endfor$
$endif$
$endif$

I wonder if we could harmonize the natbib and biblatex behaviors by removing the biblio-style option in natbib

DunLug:
@cderv I didn't use the cite package and I don't have enough knowledge about it.
However, the \cite command doesn't need the cite package to be included so it doesn't conflict with using biblatex or natbib packages.
If the cite package is relevant in the IEEE documents redaction, we should probably modify the code again to let the user choose between natbib, biblatex and cite options. This could however forbid the merging for the moment.

cderv:
> However, the \cite command doesn't need the cite package to be included so it doesn't conflict with using biblatex or natbib packages.

I just tested by activating the provided option `with_cite` and it conflicts with biblatex if included by default. `with_cite` is part of the template already so if anyone is using it (i.e setting to TRUE), I think it will break.

swhaat:
hi @DunLug and @cderv.. i had to use this template again so i thought i give it a try again..

i tried the `with_cite` option with the default template, and it does not actually work as it is described on the [template](https://github.com/rstudio/rticles/blob/main/inst/rmarkdown/templates/ieee/resources/template.tex#L95), references are listed as they were listed before, by name, and not as the documetnation says, with numbers e.g.  `[1],[2]`.

So it either has been broken all along or it wasnt used so much that people realized it, either case, to me, it makes more sense to try and remove this option as it, in reality, should not break anyones files, as the results are not what they are supposed to be.

PR #272, opened 2020-01-29
Removing formatted_doi field from JOSS template.

arfon:
This is a small change to the [JOSS](https://joss.theoj.org/) paper template to remove the `formatted_doi` field which causes production issues for JOSS and is not actually a field that we [ask for in our template](https://joss.readthedocs.io/en/latest/submitting.html#example-paper-and-bibliography).

More specifically, by including this field, the value associated overrides the parameter when it's being passed in by the JOSS Whedon bot.

This issue arose in https://github.com/openjournals/joss-reviews/issues/2028#issuecomment-578242951 when the `formatted_doi` field was left blank by the submitting author @embruna. 

Fixes https://github.com/openjournals/joss/issues/672


arfon:
Also, I didn't sign the rstudio CLA as I consider this a 'typo-level' fix.

PR #274, opened 2020-02-12
Update Copernicus template

nuest:
Also some minor fixes


nuest:
@yihui Apologies for the delay, I need to reduce my number of watched repos on GH...

Fixed in the latest commit.

PR #286, opened 2020-04-02
Modify rjournal_article template to allow more than one affiliation per author

RLumSK:
## Motivation

I realized that the `rjournal_article` template does not support multiple affiliations per author, which are, however, sometimes needed (and requested by the affiliations you are attached to).
On top of it, I realized that the fields `url` and `orcid` were not yet supported.
My contribution enables multi-affiliation support for this template and addresses #287 so that the
the output will be more consistent with the RJournal author's instructions (I am sorry, it accidentally ended up in the same pull request). 

The commit includes the following modifications: 

## Changes
* `resources/template.tex` now knows how to handle multiple affiliations, while the default behavior of the template did not change. In other words, if the author has only one affiliation and enters the information as before, the PDF output renders as before. 
* I added fields for `url` an `orcid`
* I updated the `skeleton.Rmd` to have an example for multiple affiliations
* I updated the `DESCRIPTION` and added me as a contributor (more is not justified given the minor nature of my contribution)
* I added a line to `NEWS.md`
* I address issue #287 

## Forms
- [X] Unless you have done it in any other RStudio's projects before, please sign the [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). You can send the signed copy to <jj@rstudio.com>.

- [ ] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

_Not applicable._

- [ ] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname_article/resources/template.tex`.

_Not applicable, but the original template was updated._

- [ ] Add a skeleton article `inst/rmarkdown/templates/journalname_article/skeleton/skeleton.Rmd`.

_Not applicable, but the original skeleton was updated._

- [ ] Add a description of the template `inst/rmarkdown/templates/journalname_article/template.yaml`.

_Not applicable._

- [ ] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. Please keep the number of new files absolutely minimal, and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include one hundred items in it without using all of them).

_Not applicable._

- [ ] Update Rd and namespace (could be done by `devtools::document()`).

_Not applicable._

- [X] Update NEWS.

_Not applicable._

- [ ] Update README with a link to the newly supported journal.

_Not applicable._

- [ ] Add a test to `tests/testit/test-formats.R`.

_Not applicable, the tests do not change, but I ran all tests to make sure that I did not break something._


inst/rmarkdown/templates/rjournal_article/resources/template.tex
```diff
@@ -18,10 +18,19 @@
 
 $for(author)$
 \address{%
-$author.name$\\
-$author.affiliation$\\
-$for(author.address)$$author.address$$sep$\\ $endfor$\\
+$author.name$
+$for(author.affiliation/pairs)$
```
cderv, line 22: @RLumSK is there something specific to activate with pandoc. This causes an error with me and I don't know why... 🤔 
I get this error 

```
"template" (line 22, column 24):
unexpected "/"
expecting "." or ")$"
```

It happens on my windows 10 computer, but also in Ubuntu 20 (using WSL)

Not sure why... Pandoc Manual explains clearly this but I can't use (https://pandoc.org/MANUAL.html#pipes)

yihui: I have never used pipes in templates myself.

I checked the release notes of Pandoc, and it seems that the pipes in templates were introduced in v2.8: https://github.com/jgm/pandoc/releases/tag/2.8 (it was originally called "filters" but renamed to "pipes" later: https://github.com/jgm/pandoc/releases/tag/2.9)

cderv: Oh Ok. That would explain why it passes the CI checks. 
https://github.com/rstudio/rticles/blob/ec5e82b21c89ecb2d62ea935b1a3104fd0e48375/.travis.yml#L7

That means it does not work on RStudio IDE with the shipped version. I think we want it to work.

cderv: Yep that works with last pandoc version 2.10.1. Thank you! 
Maybe we should make the CI test RStudio version and last version ?

R/rjournal_article.R
```diff
@@ -6,6 +6,30 @@
...
+  writeLines(text = Rmd_file_text, con =  Rmd_file_name)
+
+  ## correct bib-filename to match the name of the Rmd-file
+  file.rename(from = bib_file_name, to = paste0(xfun::sans_ext(Rmd_file_name), ".bib"))
+
+  ## create R file with executable R code as requested
+  knitr::purl(Rmd_file_name, documentation = 1)
+
+  ##remove objects
+  suppressWarnings(rm(Rmd_file_name, Rmd_file_text, bib_file_name))
```
cderv, line 31: @RLumSK , I have think about this today.

IMO It seems a bit too much to do all that in the format function. At least this should be done with `rmarkdown` using one of the `pre_*` or `post_*` function in the format. This would let the function access easily the input file name and the metadata, which we want both. (the `bibliography` information is in the metadata field. 

So I agree that the required R file can be generated automatically from the Rmd file using `purl`. I think this could be done in the `post_knit` step where you can easily access the input file name to do that. 
* About the Rmd file name and bib filename, I think we can also only add to the `post_knit` step because in there we have access to both the input Rmd file name, and the metadata field. 
I think we could be less impactful on the user's files by checking and warning if they are not the same. I am not sure about renaming ourself the file of the user is a good idea. I would just remind the rule of the requirements (improving the documentation is also a good way to do that). 

What do you think ? 
Are you willing to try on that or do you want me to take the lead on this ? 


Note: 
* I think The bibliography should be added in the yaml header and not at the end of the Rmd file. That may have mislead you. As we can see it is correctly taken into account in the template
https://github.com/rstudio/rticles/blob/ec5e82b21c89ecb2d62ea935b1a3104fd0e48375/inst/rmarkdown/templates/rjournal_article/resources/template.tex#L15-L17

inst/rmarkdown/templates/rjournal_article/resources/template.tex
```diff
@@ -18,10 +18,19 @@
...
+  $else$
+   \\$author.affiliation$\\
+  $endif$
+$endfor$
+$for(author.address)$$author.address$$sep$ \\$endfor$%
 }
-$if(author.email)$$author.email$$endif$
+$if(author.url)$\\$author.url$$endif$%
+$if(author.orcid)$\\\textit{ORCiD: \href{https://orcid.org/$author.orcid$}{$author.orcid$}}$endif$%
+$if(author.email)$\\$author.email$$endif$
```
cderv, line 34: Per the `RJtemplate.tex` file from RJournal, this should be inside `\adress{}`

From the template tex file of R journal: 
```tex
\address{Author Three\\
  Affiliation\\
  Address\\
  Country\\
  (ORCiD if desired)\\
  \email{author3@work}}
```

* For email, we should also use like the template : `\email`. 

* Url does not seem to be part of the RJournal template. Do you know it this can really be added ?

inst/rmarkdown/templates/rjournal_article/resources/template.tex
yihui: I changed the version in 1c5dfcd7a1e2e90b7bec87c9c1b074d340e4f87b because some users reported problems with the `cslreferences` environment, which was created by Pandoc 2.8.

I wonder if there is a way to avoid using pipes here since the current stable release of RStudio ships Pandoc 2.7.x.

R/rjournal_article.R
RLumSK: @cderv Sorry for the late response I was away from the internet. I am not sure whether I understand the comments correctly. 

## BIB-file renaming

The renaming of the BIB-file is done because it is the requirement by the R Journal (https://journal.r-project.org/share/author-guide.pdf): 

> Create a minimal ‘filename.bib’ BibTeX file with only the entries you need, without unnecessary fields such as abstracts, and add \bibliography{filename} at the end of ‘filename.tex’, where filename is the same file name created above.

So the author does not have much choice. Whatever the name of the BIB-file is, in the end, the journal wants to have it named similar to the TEX-file. Since the template is created with `'rticles'` it should be, however, not a problem. Of course, we could at this to the documentation, but to me, it is somehow odd to modify things manually when I use `'rticles'` (same goes for checking and warning). 

Nevertheless, I can understand that the authors want to use the YAML-header to define the BIB-file because they are used to it and maybe someone uses the template and then copy & paste an already defined BIB-file or something like this. So my suggestion would be as follows: 

1. I add it to the YAML-header as requested. 
2. The moment the PDF is built, it creates a subfolder 'Submission' with all the files you would need using the correct naming scheme. So basically: Ready for submission; just in a different folder. This way, it does not touch the user's files but creates a new subfolder (writing on the hard drive, but `'rmarkdown'` is doing it anyway).

What do you think?

## R-file

> So I agree that the required R file can be generated automatically from the Rmd file using purl. I think this could be done in the post_knit step where you can easily access the input file name to do that.

Sorry, I did not get it. What do you mean with "input file name". The name has to be the same as the TEX-file and it is generated from the RMD-file. Why do we want to have an option for another file name? Probably I am overlooking something.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: @cderv Thanks for double-checking.

1. I modified the address braces as requested. However, the `\email{}` does not work and causes odd LaTeX errors. Since, however, the e-mail is automatically formated correctly already, I presume that there is a conflict with some LaTeX package that was updated after the R Journal template was created. Because I already get the needed `\href{mailto:...}` tag even without using `\email{}`.

2. Regarding the URL: This is a very good question. I have seen this in published R Journal articles (example: https://journal.r-project.org/archive/2019/RJ-2019-049/RJ-2019-049.pdf). However, as you wrote, it is not part of the official template. Should I remove it? It does not seem to be a problem and in the example, one author is a former editor of the journal, so I guess it is ok. But of course, if you prefer, I remove this.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: @yihui @cderv Sorry, I was not aware of the different version shipped with RStudio. However, to address the question of whether this can be done without pipes: Yes, we can modify the template out of R on the fly. This was my first idea,  but it appeared too ugly, so I came up with the pipe solution. However, I was not aware of the RStudio pandoc version issue, but maybe it gets updated also soon in RStudio(?).

cderv: We need to provide a working template for the current RStudio version.  One solution would be to only use the `for` loop without the pipe, and that would mean affiliation would need to be defined in the Yaml as a list, even for one author.

```yaml
author:
  - name: Author One
    affiliation: 
      - name : Affiliation
      - address:
        - line 1
        - line 2
    url: https://journal.r-project.org
    orcid: 0000-0002-9079-593X
    email:  author1@work
  - name: Author Two
    url: https://journal.r-project.org
    email: author2@work
    orcid: 0000-0002-9079-593X
    affiliation: 
      - name: Affiliation 1
      - address:
        - line 1 affiliation 1
        - line 2 affiliation 1
      - name: Affiliation 2
      - address:
        - line 1 affiliation 2
        - line 2 affiliation 2
```
but this would be a breaking change because the former syntax would not work anymore for only one author. 

We could also add new `affiliation2` and `adress2` fields in the template, so that an author can add another one. This would make it less clean but would make possible the old syntax. And we can change to pandoc pipes for variable when pandocs upgrade in RStudio - in RStudio 1.4, it seems it upgrades to pandoc 2.9.2.1 from the rstudio daily build.

```
$author.affiliation$\\
$for(author.address)$$author.address$$sep$\\ $endfor$\\
}
$author.affiliation2$\\
$for(author.address2)$$author.address2$$sep$\\ $endfor$\\
}
```

How many affiliations one author could be part of for such an article ? Is two enough for most of the case ? Would could make it three without trouble.

R/rjournal_article.R
cderv: > The moment the PDF is built, it creates a subfolder 'Submission' with all the files you would need using the correct naming scheme. So basically: Ready for submission; just in a different folder. This way, it does not touch the user's files but creates a new subfolder (writing on the hard drive, but 'rmarkdown' is doing it anyway).

I like this idea. 

To sum up the situation about the bib file.

* R journal requirement is to have the .bib file name like the .tex file.
* The `rjournal_article` has an example bib file called `RJreferences.bib` in its skeleton folder. It can't be renamed when copied to the user folder so it needs to be renamed afterwards manually or automatically. 
* Let's not that the `template.tex` already support the bibliography filename to be set in the yaml header (`bibliography: `) and I think it should be used instead of `\bibliography{RJreferences}` in the Rmd file. 
* The .tex filename will be the one 

What we can do : 
1. Check and warn the user if the bib filename provide in yaml is not the same as the tex name rendered, and let the user correct
2. Rename the bib file in the user's place - the original file would be renamed
3. Copy and rename the bib file in another submission folder - the original file would still be there but unuseful at the end

I would think 1 is enough as it is a good thing that a user is aware of the requirements of its submissions. Maybe 2 is what we should do if we want to do it ourself. 3 is in the middle but it can always have side effect when we begin to play with folder. 

@RLumSK, about the input file names, what I mean is that we can use `rmarkdown` included `pre_*` or `post_*` features in the `output_format` to have access to what we need, and do something safer that a `list.files` to get access to those names. 

For examples, if someone sets `output_file` in `rmarkdown::render`, all the files should be named accordingly. This value could be accessed in the post processor for example. Currently, your implementation won't allow that. 
I can move the logic where I think it is best if you want, once we have decided what is best to do.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: Maybe we should go for the first list solution. This was also something I had tried, but because it would have broken the old syntax, I did not go for it (again, I was not aware of the RStudio issue). Nevertheless, it is an elegant and still logic way and more important: It probably needs no further update in the near future. If ok for you, I can change the PR accordingly.

R/rjournal_article.R
RLumSK: Ok to me, the best solution appears to be: 

* We do not touch the user's files, and the bibliography uses the YAML-header (with a default, but the user can do whatever wanted)
* We use the `post_*` feature to create the wanted files 
* We also use the `post_*` feature to define the output folder with default being `submission`. This way the user can even create something like 'Initial submission', 'Revision' whatsoever. 
* I have to update the documentation

If ok, I would implement it as detailed above and then we can see whether we need additional changes.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
cderv: Thanks for the example ! If it already has been published with one, we can assume it can be done. 👍 
Maybe we should add comment in the skeleton to precise which field are optional (`url` and `orcid` from what I understand.)

About `\email`,
* Currently we get `\href{mailto:author1@work}{\nolinkurl{author1@work}}` per pandoc writer. 
* the Rjournal define `\email` like this : 
https://github.com/rstudio/rticles/blob/ec5e82b21c89ecb2d62ea935b1a3104fd0e48375/inst/rmarkdown/templates/rjournal_article/skeleton/RJournal.sty#L231

The later may not work correctly with pandoc without escaping the `@`
```yaml
email:  author1\@work
```

I was just wondering the difference. I guess we can leave as is, as nobody said this wasn't accepted before ?

inst/rmarkdown/templates/rjournal_article/resources/template.tex
```diff
@@ -18,10 +18,19 @@
 
 $for(author)$
```
cderv, line 19: I reran everything and I think there is too many newlines added. There should not be one before the url. Only between the two authors.

Now: 
image

Before : 
image

R/rjournal_article.R
cderv: Let's do it this way ! 

About those `pre_* ` and `post_*` steps in `?rmarkdown::output_format`, I think you have access to the Rmd file name in the `pre_knit` or `post_knit` step - maybe `purl` should be done there, then the file renamed according to `output_file` later on on the `post_processor` function.

For the output folder, not sure it is easy to make customisable. I let you try. If modifying the user's file is simpler, you'll tell me.

Thanks!

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: @cderv Ah ok, thanks for the hint. So it is pandoc which acts smart. I thought it is a LaTeX package (e.g., 'url') that parses the e-mail addresses.  Anyway, yes, then we should leave it as it is. The entire idea of the command is to have this `\href{mailto:...}`.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: @cderv Hm, I know. This was intentional. Otherwise, it does not really format well with two affiliations. 

The idea: Two affiliations (for instance) the author may have to add because it is a request. However, it does not make much sense to have two different e-mail addresses listed (or at least it is not a requirement usually coming from the affiliation, as long as they are identified). So I added the additional line to make clear that it is valid for the two affiliations (same for ORCID and URL). Does it make somehow sense to you?

inst/rmarkdown/templates/rjournal_article/resources/template.tex
cderv: I would go for the second until a correct pandoc versions is updated in RStudio - I don't think it is a good idea to break this, and that it worth it because I am not sure a lot of authors requires mutliple affiliations as it seems you are the first to ask for it. 
Do you think 2 is not enough for the second solution ?

inst/rmarkdown/templates/rjournal_article/resources/template.tex
cderv: Perfect. So this is resolved!

inst/rmarkdown/templates/rjournal_article/resources/template.tex
cderv: I am just not sure it is ok with the R Journal requirements - I don't want to suggest it is if we are not sure.

I tried to find examples in the last R Journal issue, but it seems each author has only one affiliation (sometimes several departments but same adress)

Do you have an example of R Journal that would let us believe this is ok ?

RLumSK: @cderv You are right, I did not find any example for these in published manuscripts either (I searched before I implemented this). However, as written initially, as an author with more than one affiliation, you do not have much of a choice. Of course, the journal could say "No we don't like this", but at the same time, they know that the pressure in academia regarding the reporting of results and the correct association of affiliation is very high. Means, I am certain that a manuscript will not become rejected because the author has to add two (or more) affiliations. 

However, more important is that if there is only one affiliation: Nothing will change. I guess this is a strong argument. In the worst case (the additional line is not wanted), it takes two seconds in the copyediting process to remove this line.

cderv: > However, more important is that if there is only one affiliation: Nothing will change.

In the example first author has only one affiliation and there is a new line before the url, ORCID and email. So it seems to me this has changed.

> So I added the additional line to make clear that it is valid for the two affiliations (same for ORCID and URL). Does it make somehow sense to you?

And if you group all in one block, would that not be enough ? It would be pretty clear that the email, orcid and url are for both affiliations.

RLumSK: @cderv Ok, I will try this and see how it works out.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: Ok, then we go for solution 2.

RLumSK: Done. Please double-check.

R/rjournal_article.R
RLumSK: @cderv I decided to hook this up on the post_processing function. I played with a couple of options, but the way it is now implemented was the only reasonable way. First I had the crazy idea that the output_dir can be actually defined as input, but then I realised that the argument `output_dir` is a dead-end in `rmarkdown::pdf_document()`. It calls an internal function named `pdf_pre_processor()` but this function does nothing with the argument. 

What I still don't like is the way I identify the figures before I copy them. However, for the moment I cannot come up with a better solution.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
RLumSK: Resolved, please check.

RLumSK:
I guess the NEWS still need an update, but first I wait for your feedback.

inst/rmarkdown/templates/rjournal_article/resources/template.tex
cderv: That is awesome ! Thanks!

inst/rmarkdown/templates/rjournal_article/resources/template.tex
cderv: It seems fine ! I'll thorougly test that and adjust if necessary ! Thank you!

R/rjournal_article.R
cderv: Thanks a lot for having looked into this !

I feel that creating a submission folder may not be the safest way - you're right that the figures are not easy to deal with. From past experience with Rmarkdown, playing with the files and dirs could lead to unforseen issue. 

I think here we can go back one step to be more aligned with other **rticles** formats and what we want to support in this package

- renaming the tex file and the R file the same are ok to do because we create those anyway
- about the bibliography, I think it is the user responsability to provide one and we can't assume the default. If we can document in skeleton that it needs to be renamed, and also warn the user if not done. it seems enough. 
I know you said you find that odd to do the renaming yourself or have a warning, but as any other format, the bib file is provided by the user as one should fill it anyway - so it is one step to "save as" the file with another name. Also we can assume that the user should be familiar with the journal requirements on the bib expectation. 
At the end, **rticles**  is only here to help at its best, and if one renaming is left out for the user to do, it seems ok.
- For the figure, I'll check that it works as expected. 

Are you ok if I take it from here and modify your post processor step accordingly ?

RLumSK: @cderv Sure, thanks, feel free to take it from here. 

For the bibliography (sorry again some arguments):

* Now the YAML-field `bibliography` defines everything and the user can do whatever they want with the file, and it is already consistent with the other formats in `'rticles'`, right?
* Creating the submission folder would be only unsafe if the folder already exists in the folder, for instance, for all users who updated the package and rebuild, therefore, the exception in the code. Still, even then, it would not overwrite other user files (but maybe I am overlooking something here). In either case (update or new template): If `output_dir` in the YAML-header does not exist, nothing happens. 
* There is another point to it: Before @phargarten2 raised the issue, frankly, I was not aware of it, because it does not feel logic (but maybe it is). What matters for LaTeX is that the entry in the TEX-file matches the BIB-file. Since the default was ok, it never crossed my mind that I should rename the file. Still, you have a strong point: the user should be aware of the author's guidelines, and I agree that `'rticles'` cannot and shall not take that away from the author. However, I also believe that it saves authors a lot of time if the format issues are handled all correctly (therefore my contribution). This would allow authors to focus on the essential aspects of the author's guidelines regarding science.
* A warning message is possible. But shouldn't a warning flag an exception? Here, the warning would become the standard message, shown to every user using the template and hitting the `Knit` button the first time. Except for one case that the authors read everything super carefully, and the first thing done was renaming the BIB-file **and** adjusting the TEX-template. 

I found the current way is elegant (it does not mean that the code couldn't be more sophisticated) and user friendly, and it reflects functionality I would have "expected" in the first place. 

However, you have the bigger picture in mind, and so of course, the final decision is all yours. Thank in either case for your work the professional support!

yihui: I'll revisit to this PR after @cderv finishes the revision. Depending on the complexity of implementation, we might reconsider the decision on what to do for the bibliography file. Thank you @RLumSK for the contribution and discussion!

R/rjournal_article.R
```diff
@@ -1,35 +1,56 @@
...
 #'
 #' *Please note: Only one `url`, `orcid` and `email` can be provided per author.*
 #'
-#' **Other YAML fields**
+#' ## Other YAML fields
+#'
+#' | FIELD | TYPE | DESCRIPTION |
+#' | ----- | ---- | ----------- |
+#' | `bibliography` | *with default* | the BibTeX file with the reference entries |
+#' | `output_dir` | *with default* | output directory where all the files are stored ready for submission |
```
RLumSK, line 52: The `output_dir` is not anymore part of the YAML-header after the new commit.

cderv: Good catch ! I am still checking everything - if you see anything else do not hesitate.

R/rjournal_article.R
```diff
@@ -1,30 +1,126 @@
...
+#' | FIELD | TYPE | DESCRIPTION |
+#' | ----- | ---- | ----------- |
+#' | `bibliography` | *with default* | the BibTeX file with the reference entries |
+#'
 #' @param ...,citation_package Arguments to \code{rmarkdown::pdf_document}.
+#'
+#' @md
 #' @export
-rjournal_article <- function(..., citation_package = 'natbib') {
+rjournal_article <- function(..., keep_tex = TRUE, citation_package = 'natbib') {
```
RLumSK, line 57: The `keep_tex` has no documentation entry (throws a warning if you check the package)

cderv: `pdf_document` params where not inherited here as in other **rticles** function. It is fixed

R/rjournal_article.R
cderv: I resolve this discussion and we can continue in comment feed.

cderv:
I now have revised the implemention to fix a few issues and refactor some code.  Here, we only rename R file and tex file and we only document and warn for bibliography naming. Documentation of the format has also been improved. Figures are now kept in the `*_files` that rmarkdown creates as any other format and I believe pdf file can be generated from tex file. 

I like the md documenting for roxygen so I kept that for this one even if it is the first function to be documented this way currently. Thanks for suggesting that ! 

@yihui You can now have a look at it and I think we have now a new good base to rethink this bib renaming thing. 

Let's not that currently (as in _master_) the `rticles::rjounal_article` format do not support setting `output_dir` in `rmarkdown::render`. I did not correct that yet in this PR.

RLumSK:
@cderv Very nice work, indeed. Interesting to see how you solved the issues differently. I also played with the `post_knit` but it did not work out (of course this was because I did this file copying into a separate folder and when the function runs not all files are available). 

The only thing we may want to keep in mind: I am not sure (nothing is written in the author's guidelines) whether they allow subfolders for the submission or prefer a flat-file structure (i.e. no subfolder like, e.g., Elsevier usually requests).

cderv:
> The only thing we may want to keep in mind: I am not sure (nothing is written in the author's guidelines) whether they allow subfolders for the submission or prefer a flat-file structure (i.e. no subfolder like, e.g., Elsevier usually requests).

You are talking about the figures ? I asked myself the same question and found nothing in the author's guidelines either.

cderv:
Another question I add: can the `\author{}` in Tex file be on several line or not ? 
If so, the post processing of authors should be modified accordingly. Currently it handles only if on one line in the tex file.

RLumSK:
> > The only thing we may want to keep in mind: I am not sure (nothing is written in the author's guidelines) whether they allow subfolders for the submission or prefer a flat-file structure (i.e. no subfolder like, e.g., Elsevier usually requests).
> 
> You are talking about the figures ? I asked myself the same question and found nothing in the author's guidelines either.

Yes, I meant the figures. I suggest to leave it as it is for the moment. If no one raises that issue again, it was ok.

RLumSK:
> Another question I had: can the `\author{}` in Tex file be on several line or not ?
> If so, the post processing of authors should be modified accordingly. Currently it handles only if on one line in the tex file.

I was inclined to write no because the template uses

`\author{by $for(author)$$author.name$$sep$, $endfor$}`

However, I just tried with to create a document with six authors and "voilà", there was a linebreak and the replacement did not work anymore the way it should. Means, you are right, we have to consider a linebreak.

RLumSK:
@cderv One last comment for today. Now where the warning is shown for the bibliography, there is one problem, at least when I use *RStudio*: If the markdown process does not show an error the panel window (in *RStudio*) switches immediately back from the R Markdown Panel to the Console. This happens all so fast that it is impossible to notice a warning message. If I back and check, of course, the warning is there. But imagine the user never does something really wrong, never causes an error message, there would be no reason to look into the r markdown terminal feedback. So maybe the warning should become a `stop()` instead?

cderv:
@RLumSK Thank you for your input code. I prefered to handled it differently while testing more edge cases. Hope you don't mind! 

About the warning, it appears in the R Markdown pane indeed as any other warnings for the R code. So if there is a warning in the R code, and it is not captured in the Rmd output, it will also be printed inside this Rmarkdown pane. 
The user has already a comment in the skeleton to  remind to change the name 
https://github.com/rstudio/rticles/blob/51d27bf4c22e58d5f6939ddb5c28be7b1c34a5ae/inst/rmarkdown/templates/rjournal_article/skeleton/skeleton.Rmd#L36-L38
This and the warning in addition should be enough - I think adding a `stop()` is too much.

If we wan't to improve that, we still have the solution of (copying+) renaming the bib file and modifying the file accordingly. As @yihui said, he will look into this and maybe reconsider the current choices.

RLumSK:
> @RLumSK Thank you for your input code. I prefered to handled it differently while testing more edge cases. Hope you don't mind!

Sure, no problem, it was just a suggestion. Looks nice the way you did it.

R/rjournal_article.R
```diff
@@ -1,30 +1,127 @@
...
+#' ## Other YAML fields
+#'
+#' | FIELD | TYPE | DESCRIPTION |
+#' | ----- | ---- | ----------- |
+#' | `bibliography` | *with default* | the BibTeX file with the reference entries |
+#'
+#' @inheritParams rmarkdown::pdf_document
+#' @param ... Arguments to `rmarkdown::pdf_document.
+#'
+#' @md
```
yihui, line 56: I feel it may be a better idea to put the main documentation in `skeleton.Rmd`, because I'm not sure if authors would actually read the help page. What do you think?

RLumSK: Personally I think that it is cleaner to have it in the documentation of the function (as it is). The documentation in the RMD-file for the R Journal is already neat and intelligible. I would not add too much more here. My understanding is that `'rticles'` supports the writing process, so less technical details might be better.

cderv: The part **About this format and the R Journal requirements** is already in the skeleton. 
FOr the **The author field in the YAML header** section, I'll add a mention in the skeleton to consult the documentation.

cderv:
Thanks a lot @RLumSK  for this PR ! Great work and discussions !

RLumSK:
@cderv Likewise! I also carefully locked into your changes and I hope next time I can meet your expectations from the start. Anyway, it was nice and sincerely exciting work.

PR #304, opened 2020-08-04
Add the contributors to the list of journal templates in README

### LINKED ISSUE #302: Add the contributors to the list of journal templates in README
Search the merged PRs to find out the contributors to existing templates, so we know whom to ping when users have questions or issues with a particular template.

cderv:
This will close #302

I went through all the template and their git history. Several things in this PR:

* I added the first contributor mainly to templates that were contributed by the community. Some other templates are from JJ or Hadley, so I did not include their name. Should we include them ? Or by default, if no name, it is us ? 
* Some template where contributed a long time ago, without the PR. I could have add the commit, but seems a bit too much, so I just added the handle. 
* Some urls were corrected
* Some templates were missing in the list


README.md
```diff
@@ -21,55 +21,60 @@ The **rticles** package provides a suite of custom [R Markdown](http://rmarkdown
 
 - [ACM](http://www.acm.org/) articles
 
-- [ACS](http://pubs.acs.org/) articles
+- [ACS](http://pubs.acs.org/) articles (contributed by @yufree)
```
yihui, line 24: I forgot to mention that besides the full journal names, it might be a good idea to also add the output format name in the list like:

```suggestion
- `rticles::acs_article`: [ACS](http://pubs.acs.org/) articles (contributed by @yufree)
```

cderv: I add the same idea before I see this comment but I put it at the end without `rticles::`
https://github.com/rstudio/rticles/pull/304/commits/62cfd1c191736d4255ebd1d91c3ed09e3559f7c9

Do you prefer it at the beggining as you show here ?

README.md
```diff
@@ -19,57 +19,63 @@ remotes::install_github("rstudio/rticles")
...
+- [CTeX](https://ctan.org/pkg/ctex) documents - `ctex()`
 
-- [JSS](http://www.jstatsoft.org/) articles
+- [Elsevier](https://www.elsevier.com) journal submissions - `elsevier_article()`
 
-- [JOSS](http://joss.theoj.org/) and [JOSE](https://jose.theoj.org/) articles
+- [Frontiers](https://www.frontiersin.org/) articles (contributed by @zkamvar) - `frontiers_article()`
 
-- [MDPI](http://www.mdpi.com) journal submissions
+- [IEEE Transaction](http://www.ieee.org/publications_standards/publications/authors/author_templates.html) journal submissions (contributed by @Emaasit, @coatless, @espinielli, @nathanweeks , @DunLug) - `ieee_article()`
```
cderv, line 48: IEEE format does not have one main contributor but rather several. I had all the relevant one 
https://github.com/rstudio/rticles/commits/master/inst/rmarkdown/templates/ieee_article

Is this ok ? the list is rather long so I am wondering... 🤔

yihui: That's ok.

README.md
yihui: I don't have a preference. It's fine to put it at the end.

Issue #254, opened 2019-10-20
Suggested improvements for JSS support


statibk:
By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('rticles')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/rticles')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

---

After authoring my first own rticles-based JSS submission, I wanted to thank you for providing Rmd support for JSS and suggest some small improvements.

**Continuation prompts**

At the moment rticles employs either `R+ ` (for evaluated code) or `R> ` (for non-evaluated code) as the continution prompt but `+  ` would be requested by JSS. As an example,

    ```{r}
    f(a = 1,
      b = 2)
    ```

is currently displayed as follows when combined with `eval=TRUE`

    R> f(a = 1,
    R+   b = 2)

and with `eval=FALSE` as

    R> f(a = 1,
    R>   b = 2)

but

    R> f(a = 1,
    +    b = 2)

would be desired.

**Address formatting**

In `skeleton.Rmd` a simple `>` continuation is used for the `address`. However, to force the right line breaks `|` continuation would be better. I would suggest instead of

    address: >
      First line
      Second line

to use:

    address: |
      | Department/Faculty/etc.
      | Street Number
      | ZIP City, Country

**Inline code formatting**

For code in Markdown it is ok to use backtick syntax, so that `\code{...}` is typically not necessary. Hence I it is ok to remove `\code{print("abc")}` from `skeleton.Rmd`, I think.

**Escaping code/math in section titles**

How can code/math formatting in section titles be escaped? For example, when I want in LaTeX

    \section[R code]{\proglang{R} code}

Can this be done in Markdown/Pandoc? If so, an example in `skeleton.Rmd` might help.

Thanks for all your efforts regarding JSS support, these are much appreciated!
Achim

yihui:
[...]

PR #310, opened 2020-08-10
add a lua script to include short titles on headers

cderv:
This contributes to #254 by trying to use a lua filter to generate short title from headers. 

to get from 
```md
# \proglang{R} code {short-title="R code" #id}
```
this latex syntax
```tex
\section[R code]{\proglang{R} code}\label{id}
``` 

The approach here is to detect headers with `short-title` attributes and transform them to latex code directly to not let pandoc do it. (as it does not support short titles for now)

This is working I think but the drawback is that we do the job of pandoc for those header, to we need to sure this works for many cases.

## needs to be done

- [x] ~Maybe add `\part` command to~ (`\part` and `\chapter` won't be used in such articles)
- [x] add `\label` after the section for referencing
- [x] Clean code when the logic will be ok
- [x] tests on several cases  
- [x] Document this new feature


R/jss_article.R
```diff
@@ -39,6 +39,16 @@ jss_article <- function(..., keep_tex = TRUE, citation_package = 'natbib') {
     )
   }
 
+  pre <- base$pre_processor
+  base$pre_processor <- function(metadata, input_file, runtime,
```
yihui, line 43: I'd probably add a `pandoc_args` argument to `jss_article()`, and pass `--lua-filter` to it, instead of using a pre-processor.

inst/rmd/lua/short-title.lua
```diff
@@ -0,0 +1,17 @@
+function Header(el)
+  local options = el.attributes['short-title']
+  if options == nil then return nil end
```
yihui, line 3: We need to check the JSS requirements (https://www.jstatsoft.org/pages/view/authors) and see if plain short titles are always required. If so, we need to generate a plain-text version (no formatting commands) of the header and use the text as the short title.

R/jss_article.R
cderv: Oh yes great idea! Thank you !

inst/rmd/lua/short-title.lua
cderv: Only `\section` using special latex syntax use a short title like
```tex
\section[Using BibTeX]{Using \textsc{Bib}{\TeX}} \label{app:bibtex}
```
However, they often use `\label` (not mandatory too).  Pandoc adds a label for each header. We could also do that, but we need to generate an identifier from the text.

cderv: `\label` has been added based ton the header identifier built by Pandoc itself.

cderv: However, I don't really know if header identifier is always ok. In pandoc a special `toLabel` function is used to create this `\label{}` part, that calls `stringToLaTeX` that aims at "escape things as needed for LaTeX"
So I guess, the header identifier won't always be ok. 🤔 

As a fail safe, we could add a label attributes - if provided we use that as label ? Or maybe it is was the `{#id}` is for.

cderv: > As a fail safe, we could add a label attributes - if provided we use that as label ? Or maybe it is was the {#id} is for. I'll check.

Yes - it works. 
```md
# \proglang{R} code {short-title="R code" #r-code}
```
will be correctly transformed to 
```tex
\section[R code]{\proglang{R} code}\label{r-code}
```

cderv:
Do you think we need to to add the latex command `hypertarget` like pandoc ? See https://github.com/rstudio/rticles/issues/254#issuecomment-671389817
or can't we omit this for JSS article ?

R/jss_article.R
```diff
@@ -5,12 +5,17 @@
 #' @inheritParams rmarkdown::pdf_document
 #' @param ... Arguments to \code{rmarkdown::pdf_document}
 #' @export
-jss_article <- function(..., keep_tex = TRUE, citation_package = 'natbib') {
+jss_article <- function(..., keep_tex = TRUE, citation_package = 'natbib',
+                        pandoc_args = NULL) {
```
yihui, line 9: When possible, I write all arguments on one line; if not possible, I use this style:

```suggestion
jss_article <- function(
  ..., keep_tex = TRUE, citation_package = 'natbib',
  pandoc_args = NULL
) {
```

R/jss_article.R
```diff
@@ -5,12 +5,17 @@
...
 
   rmarkdown::pandoc_available('2.2', TRUE)
 
+  lua_filter <- c(pandoc_args,
```
yihui, line 13:
```suggestion
  pandoc_args <- c(pandoc_args,
```

R/jss_article.R
```diff
@@ -5,12 +5,17 @@
...
+                  c("--lua-filter", pkg_file("rmarkdown", "lua", "short-title.lua")))
+
   base <- pdf_document_format(
-    "jss_article", keep_tex = keep_tex, citation_package = citation_package, ...
+    "jss_article", keep_tex = keep_tex, citation_package = citation_package,
+    pandoc_args = lua_filter, ...
```
yihui, line 18:
```suggestion
    pandoc_args = pandoc_args, ...
```

R/utils.R
```diff
@@ -1,6 +1,6 @@
 find_resource <- function(template, file = 'template.tex') {
-  res <- system.file(
-    "rmarkdown", "templates", template, "resources", file, package = "rticles"
+  res <- pkg_file(
+    "rmarkdown", "templates", template, "resources", file
   )
```
yihui, line 4: Since it fits one line (less than 80 chars), I'd merge the three lines:

```suggestion
  res <- pkg_file("rmarkdown", "templates", template, "resources", file)
```

R/utils.R
```diff
@@ -65,3 +65,7 @@ get_list_element <- function(x, names) {
   for (i in names[seq_len(n - 1)]) if (!is.list(x <- x[[i]])) return()
   x[[names[n]]]
 }
+
+pkg_file <- function(...) {
+  system.file(..., package = "rticles")
+}
```
yihui, line 71: That's great! In the future, I think you'll have several more chances to factor out code in our other packages.

BTW, for simple functions like this, I'd write on one line, too:

```suggestion
pkg_file <- function(...) system.file(..., package = "rticles")
```

I have this habit of "saving vertical space as much as possible" because when I read code, I don't want to scroll up and down frequently. If that doesn't make sense to you, let's follow whatever coding style that you are comfortable with.

inst/rmarkdown/templates/jss_article/skeleton/skeleton.Rmd
```diff
@@ -43,11 +43,20 @@ In general, don't use Markdown, but use the more precise LaTeX commands instead:
 
 One exception is inline code, which can be written inside a pair of backticks (i.e., using the Markdown syntax).
 
-# R code
+If you want to use LaTex commands in headers, you need to provide a `short-title`
```
yihui, line 46:
```suggestion
If you want to use LaTeX commands in headers, you need to provide a `short-title`
```

inst/rmarkdown/templates/jss_article/skeleton/skeleton.Rmd
```diff
@@ -3,9 +3,9 @@ documentclass: jss
 author:
   - name: FirstName LastName
     affiliation: University/Company
-    address: >
-      First line
-      Second line
+    address: |
+      | First line
+      | Second line
```
yihui, line 8: Are the leading pipes necessary? Probably not. Could you [double check](https://pandoc.org/MANUAL.html#metadata-blocks)?

inst/rmarkdown/templates/jss_article/skeleton/skeleton.Rmd
```diff
@@ -43,11 +43,20 @@ In general, don't use Markdown, but use the more precise LaTeX commands instead:
 
 One exception is inline code, which can be written inside a pair of backticks (i.e., using the Markdown syntax).
 
-# R code
+If you want to use LaTex commands in headers, you need to provide a `short-title`
+attribute. You can also provide a custom identifier if necessary.
+See the example of the header \ref{r-code}
```
yihui, line 48:
```suggestion
See the header of Section \ref{r-code} for example.
```

BTW, since other paragraphs are not wrapped, let's not wrap this one, either.

If you prefer hard wrapping, we can use the RStudio daily build with the visual editor, so you don't have to manually wrap the text.

inst/rmarkdown/templates/jss_article/skeleton/skeleton.Rmd
```diff
@@ -43,11 +43,20 @@ In general, don't use Markdown, but use the more precise LaTeX commands instead:
...
 Can be inserted in regular R markdown blocks.
 
 ```{r}
 x <- 1:10
 x
 ```
+
+## Features specific to \pkg{rticles} {short-title="Features specific to rticles"}
+
+* Adding short title to section is specific to \pkg{rticles} as currently unsupported by pandoc. It is powered by a pandoc lua filter.
```
yihui, line 61:
```suggestion
* Adding short titles to section headers is a feature specific to \pkg{rticles} (implemented via a Pandoc Lua filter). This feature is currently not supported by Pandoc and we will update this template if [it is officially supported in the future](https://github.com/jgm/pandoc/issues/4409).
```

inst/rmarkdown/templates/jss_article/skeleton/skeleton.Rmd
```diff
@@ -43,11 +43,20 @@ In general, don't use Markdown, but use the more precise LaTeX commands instead:
...
+* Using `\AND` syntax in author field to add author in a new line. This is a specific `jss_article` format.
```
yihui, line 62:
```suggestion
* Using the `\AND` syntax in the `author` field to add authors on a new line. This is a specific to the `rticles::jss_article` format.
```

NEWS.md
```diff
@@ -1,6 +1,8 @@
 rticles 0.15
 ---------------------------------------------------------------------
 
+- Improve `jss_article()` format: Short titles can now be provided to headers to escape code/math in section titles, continuation prompt as the correct formatting and skeleton has been updated accordingly (thanks, @statibk, #254, @Freguglia, #294)
```
yihui, line 4:
```suggestion
- Improved the `jss_article()` format: Short titles can now be provided to headers to escape code/math in section titles, the continuation prompt has been corrected (from `R+` to `+`), and the skeleton document has been updated accordingly (thanks, @statibk #254, @Freguglia #294).
```

R/jss_article.R
cderv: I can adapt to this style already in place in **rticles**. 

I used that because RStudio IDE aligns function arguments this way when I go to next line so it is practical ☺️ 
And it is coherent with https://style.tidyverse.org/ that I know of.

R/utils.R
cderv: I am used to have all the inside content of a function aligned at one indent, so I do this even for a one liner. 
But I'll see if I get used to this because I see the advantage of having less lines vertically and for this short function it makes sense to use one line.

inst/rmarkdown/templates/jss_article/skeleton/skeleton.Rmd
cderv: I blindly commit the suggestion  in #254, but I'll check that.

yihui: I just checked it. They are necessary. The meanings of the pipe after and below `address:` are different. The former is explained at https://pandoc.org/MANUAL.html#metadata-blocks and the latter means https://pandoc.org/MANUAL.html#line-blocks

cderv: Thanks ! That is good to know!

PR #297, opened 2020-07-04
Add new article template for Bioinformatics

ShixiangWang:
To contribute a new article template to this package, please make sure you have done the following things (note that `journalname_article` below is only an example name):

- [x] Unless you have done it in any other RStudio's projects before, please sign the [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). You can send the signed copy to <jj@rstudio.com>.

- [x] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [x] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname_article/resources/template.tex`.

- [x] Add a skeleton article `inst/rmarkdown/templates/journalname_article/skeleton/skeleton.Rmd`.

- [x] Add a description of the template `inst/rmarkdown/templates/journalname_article/template.yaml`.

- [x] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. Please keep the number of new files absolutely minimal, and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include one hundred items in it without using all of them).

- [x] Update Rd and namespace (could be done by `devtools::document()`).

- [x] Update NEWS.

- [x] Update README with a link to the newly supported journal.

- [x] Add a test to `tests/testit/test-formats.R`.

- [x] Add your name to the list of authors `Authors@R` in DESCRIPTION. You don't need to bump the package version in DESCRIPTION.

Lastly, please try your best to do only one thing per pull request (e.g., if you want to add two output formats, do them in two separate pull requests), and refrain from making cosmetic changes in the code base: https://yihui.name/en/2018/02/bite-sized-pull-requests/

Thank you!

***

I add this new template based on #83, so two authors are added to DESCRIPTION. Please correct my modification if there are something wrong.

This PR can close #83.

Best,
Shixiang

PR #316, opened 2020-08-19
Use only journal name for folder in templates

### LINKED ISSUE #315: Rename template folders to shortten the path
See discussion in https://github.com/rstudio/rticles/pull/297#issuecomment-675489389

One solution is to remove the `_article` part in template folder 
```
$ ls inst/rmarkdown/templates/
acm_article  arxiv_article           elsevier_article   mnras_article     rsos_article
acs_article  asa_article             frontiers_article  oup_article       rss_article
aea_article  bioinformatics_article  ieee_article       peerj_article     sage_article
agu_article  biometrics_article      joss_article       plos_article      sim_article
amq_article  copernicus_article      jss_article        pnas_article      springer_article
ams_article  ctex                    mdpi_article       rjournal_article  tf_article
```

That would require to modify each format function accordingly: This folder name is used to find the resource in `pdf_document_format()` and currently match the format name. 

## To Do 
- [x] Rename the folders
- [x] Update the format functions
- [x] Update the PR template

cderv:
This will close close #315 and resolve https://github.com/rstudio/rticles/pull/297#issuecomment-675489389 

The renaming happens in several places. Only user facing change that could be strange: when using `rmarkdown::draft`, one should use only the journal name now. 

Before template is the same as the format name
```r
rmarkdown::draft("MyArticle.Rmd", template = "acm_article", package = "rticles")
```
Now, it is just the journal name
```r
rmarkdown::draft("MyArticle.Rmd", template = "acm", package = "rticles")
```

It could be a breaking change for some users who scripted their draft creation. 

Should we stick to this solution to resolve https://github.com/rstudio/rticles/pull/297#issuecomment-675489389 ?


cderv:
>  I'm not entirely sure if this breaking change is fine

I had the same thought. Not sure either but I tend to say yes too because it is easy to fix if someone uses this. Just the template name should be changed. I believe people are creating template through RStudio. But I found a few
https://github.com/search?q=rmarkdown%3A%3Adraft+rticles+language%3AR&type=Code

This changes is also not necessary to resolve the note in #297 as we have other solution. 

> BTW, I wonder why there are changes in certain .cls and .bib files. I feel these might be CRLF changes from your GIT client on Windows.

You may be right, I will check and configure that correctly.

cderv:
> BTW, I wonder why there are changes in certain .cls and .bib files. I feel these might be CRLF changes from your GIT client on Windows.

I believe this is because we have some commited file on the DOS format. My git setup is `core.autocrlf true` as I often work with cross platform program, and it is good practice on windows. So when I renamed the path to the file, CRLF (dos format) where converted by git to LF (unix format).

I can revert but do we want DOS format file ? 

We could use `.gitattributes` in the repo to insure that it happens the same for everyone. https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings

cderv:
@yihui Two new PR are integrating what is discussed here. To be merged in due time.

Issue #329, opened 2020-10-06
JSS article throw error upon knitting


cimentadaj:
---

By filing an issue to this repo, I promise that

- [X] I have fully read the issue guide at https://yihui.name/issue/.
- [X] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('rticles')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/rticles')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [X] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

---

I'm trying to knit a fresh template of JSS and I get a `LaTeX` error. I remember knitting this template a few months ago and everything was fine. Here's the template:

```r
---
documentclass: jss
author:
  - name: FirstName LastName
    affiliation: University/Company
    # use this syntax to add text on several lines
    address: |
      | First line
      | Second line
    email: \email{name@company.com}
    url: http://rstudio.com
  - name: Second Author
    affiliation: 'Affiliation \AND'
    # To add another line, use \AND at the end of the previous one as above
  - name: Third Author
    address: |
      | Department of Statistics and Mathematics,
      | Faculty of Biosciences,
      | Universitat Autònoma de Barcelona
    affiliation: |
      | Universitat Autònoma 
      | de Barcelona
    # use a different affiliation in adress field (differently formated here)
    affiliation2: Universitat Autònoma de Barcelona
title:
  formatted: "A Capitalized Title: Something about a Package \\pkg{foo}"
  # If you use tex in the formatted title, also supply version without
  plain:     "A Capitalized Title: Something about a Package foo"
  # For running headers, if needed
  short:     "\\pkg{foo}: A Capitalized Title"
abstract: >
  The abstract of the article.
keywords:
  # at least one keyword must be supplied
  formatted: [keywords, not capitalized, "\\proglang{Java}"]
  plain:     [keywords, not capitalized, Java]
preamble: >
  \usepackage{amsmath}
output: rticles::jss_article
---

```{r, setup, include=FALSE}
options(tinytex.verbose = TRUE)
[...]

Issue #303, opened 2020-08-03
Fix travis build


cderv:
Currently build is failling https://travis-ci.org/github/rstudio/rticles/builds/714523391

It seems like it is taking too long, so travis things it fails. Maybe because of all the tex package to install. 🤔

yihui:
I guess I'm more familiar with the TeX Live stuff, so let me try to fix this issue first. If I can't figure it out, I'll leave it to you. Thanks!

yihui:
Fixed now: https://travis-ci.org/github/rstudio/rticles/builds/714523391

In case it should become a problem in the future again, we can preinstall some LaTeX packages to avoid the timeout (the time is spent on looking for missing packages and installing them):

```yaml
  - tlmgr install achemso ametsoc babel-french biblatex carlisle chemgreek crop dblfloatfix draftwatermark endfloat endnotes epsf esvect etex-pkg everypage floatflt forarray hyphen-french hyphenat jknapltx logreq marginnote mhchem mnras moreverb pdfsync psfrag rsfs seqsplit sidecap soul soulutf8 stix tabto-ltx tex-gyre totcount xypic
  - tlmgr path add
```

cderv:
This is now working. It could have been a travis issue. @yihui if you've done anything, please share.

cderv:
Awesome thank you.

PR #335, opened 2020-10-16
Support Pandoc 2.11

### LINKED ISSUE #333: Check formats and Pandoc 2.11 new citeproc
Pandoc 2.11 modifies the way citation are processed, we must check it still works with the **rticles** template. 

I know there is some changes that are missing in the templates. 

For example, Pandoc has introduced a new `cslreferences` environment that we included in our template
https://github.com/rstudio/rticles/commit/ee529ada1ee5dc144300383aa8f88364108b6679

but in Pandoc 2.11 it has now changed to `CSLreferences`
https://github.com/jgm/pandoc/commit/e0984a43a99231e72c02a0a716c8d0315de9abdf#diff-64a9e5fa4509dce2699276c9e39993b4474e2446c3e86941a2f78a379f8c4079

So we have this error: 
```
! LaTeX Error: Environment CSLReferences undefined.
```

We must find a way to support both - all the more because RStudio IDE will bundle Pandoc 2.11+ - so currently **rticles** is broken with this version.

cderv:
Pandoc 2.11 introduced this new latex environment while introducing new --citeproc argument (https://github.com/jgm/pandoc/commit/e0984a43a99231e72c02a0a716c8d0315de9abdf#diff-64a9e5fa4509dce2699276c9e39993b4474e2446c3e86941a2f78a379f8c4079)
This follows a previous change we made when Pandoc 2.8 went out (https://github.com/rstudio/rticles/commit/ee529ada1ee5dc144300383aa8f88364108b6679).

This PR aims to close #333 

See https://github.com/rstudio/rticles/issues/333#issuecomment-709940186 for some thoughts about the solution

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,13 @@ addons:
 
 before_install:
   - wget -qO- https://github.com/yihui/tinytex/raw/master/tools/download-travis-linux.sh | sh
+
+jobs:
+  include:
+    - r: release
+      name: R release / Pandoc devel
+      env: PATH=$HOME/bin:$PATH
+      before_install:
+        - wget -qO- https://github.com/rstudio/rmarkdown/raw/master/tools/install-pandoc.sh | sh
+    - r: release
+      name: R release / default Pandoc
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,14 @@ addons:
 
 before_install:
   - wget -qO- https://github.com/yihui/tinytex/raw/master/tools/download-travis-linux.sh | sh
+  - tlmgr --version
 
 jobs:
   include:
     - r: release
       name: R release / Pandoc devel
       env: PATH=$HOME/bin:$PATH
-      before_install:
+      before_script:
         - wget -qO- https://github.com/rstudio/rmarkdown/raw/master/tools/install-pandoc.sh | sh
     - r: release
       name: R release / default Pandoc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: bionic
 language: r
 cache:
   packages: true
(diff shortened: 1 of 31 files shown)


yihui (approved):
Not sure what's going on with Travis, but the changes look good to me. Thanks!

cderv:
> the changes look good to me.

Great, so we go with the easy way and not conditional on pandoc version. Templates will have the `cslreferences` environment define for nothing with Pandoc 2.11+ but I guess it does not hurt 

> Not sure what's going on with Travis

I think it fails because of newer latex that causes issues with some articles. For now we have two problematic, and one of the issue cause `pdflatex` to hang and I guess this is happening on travis. 

Some PR will fix this normally. Will see.

Otherwise, we should use an older binary release of TinyTex until it is fixed.

cderv:
You've merged the PR that should fix one of the issue (0e8335199e6f5decd8afadd07c22781ef12f6fcb) 

But as TinyTeX is cached on Travis I am not sure which version will be used at some point in master. I know it reinstalled it for checks in this PR. I am bit lost with TeX Live version in our CI workflow - it seems it is cached and so we test on older version that people could encounter.

cderv:
Oh,  is it possible that this is like last time : https://github.com/rstudio/rticles/issues/303 ? 
As all the latex package must be installed for all the templates, it takes too long for Travis ?

cderv:
Forget what I said above, in fact in fails because of #329 - so I guess that is a good sign because now our CI workflow is seeing the issue.

yihui:
You are correct. I just committed 1054e2b2418995e1ef669ab13a73eab1b79f6ba6 and let's see if it works.

cderv:
I apply the same fix as in copernicus by pinning only the Tex package that is the culprit. We may not need to pin down the all TinyTeX.

PR #339, opened 2020-10-19
install all required packages for testing rticles

cderv:
This should fix the timeout problem on travis.

List based on setdiff with tinytex::tl_pkgs() before and after running tests.


cderv:
Ok so installing latex packages explictly instead of letting tinytex guess and install solves the travis build. 

I'll merge this

yihui:
Excellent! Thanks!

Issue #298, opened 2020-07-06
Outdated AMS template


eliocamp:
The American Meteorological Society now has a new version of their LaTeX template (https://www.ametsoc.org/ams/index.cfm/publications/authors/journal-and-bams-authors/latex-author-info/) and some new guidelines (such as adding a significance statement).

I'm updating the article template but how does rticles manage these kinds of breaking changes? Do you just replace the old skeleton and template with the new one or should I create an amsV5_article, or maybe replace ams_article but preserve the old behaviour in an ams_article_old?

yufree:
As long as no AMS journal will accept the old template, I think it's fine to replace the outdated one.

eliocamp:
Part of the guidelines of the AMS is to use the `\appendcaption` instead of `\caption` for appendix figures. Right now, there is no easy way of changing the caption command, and even if there were (I've opened [an issue in knitr](https://github.com/yihui/knitr/issues/1872)) it wouldn't help in this particular case because the syntax of both commands are slightly different. 

What I'm doing now, as I dogfood myself by writing an article using this template, is to have a chunk at the top of the document that defines and sets a new plot hook. You can see it [here](https://github.com/eliocamp/rticles/blob/AMS-V5/inst/rmarkdown/templates/ams_article/skeleton/skeleton.Rmd#L100). 

It works, but is not very user-friendly. Would it be acceptable to add that custom plot hook function as a function inside rticles?

yihui:
> It works, but is not very user-friendly. Would it be acceptable to add that custom plot hook function as a function inside rticles?

That's okay. We have done that with other article formats in this package.

eliocamp:
Oh! I looked at more of the codebase. I hadn't realised that the `ams_article()` function could add hooks! That's great! I'll do that.

cderv:
@eliocamp sorry for the delay on this one. 

AMS has now updated to template V6. 
https://www.ametsoc.org/index.cfm/ams/publications/author-information/latex-author-info/

Regarding #348 I believe we should not merge V5 maybe but try V6 directly for an update ? 

I know we are way longer than we should to merge PR but I am catching up currently for next CRAN release.

@yufree @eliocamp What are your thoughts on this ? 
Is one of you still publishing to AMS ?

eliocamp:
I'm not. We moved to Climate Dynamics. I do agree that with V6 out, then it only makes sense to implement that version directly instead of V5.

yufree:
[...]

PR #340, opened 2020-10-21
AMS template.tex: Fix brace in author2

ConorIA:
As written, the authors block is never closed when there is no `\currentaddress for author 2`. I've also moved the closing $endif$ for $if(author1)$ so that we don't end up with a stray brace when no authors are specified.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rticles
 Type: Package
 Title: Article Formats for R Markdown
-Version: 0.16.1
+Version: 0.16.2
 Authors@R: c(
   person("JJ", "Allaire", role = "aut", email = "jj@rstudio.com"),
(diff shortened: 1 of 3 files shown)


cderv:
Thanks ! 

This looks ok to me. 

IMO, this part could be rewriten to use a for loop for authors like in other template. For example:  
https://github.com/rstudio/rticles/blob/6088e669160a05ecbc343c03fe7ad3ffc07c24ee/inst/rmarkdown/templates/arxiv/resources/template.tex#L17-L31

Currently, there could only be two authors for example. But maybe that is required by the template, IDK. 

ping @yufree as you are the one submitted this template in the first place. FYI.

cderv:
Thank you @ConorIA !

yufree:
Thanks for PR! 
`author1` is for the old AMS latex template to deal with two authors situation. As mentioned in #298 , AMS has released v5 template which will no longer need `author1`. However, to make the rticle AMS template work, such change is necessary. In long term, rticle AMS template will shift to new version.

cderv:
Great ! I forgot about his PR!

Issue #332, opened 2020-10-15
arxiv template - Undefined control sequence includegraphics


fouodo:
Hello,

I am trying to use the arxiv template and face to the following issue:

 Undefined control sequence.
l.136 \includegraphics

Could you please  look at this issue?

A minimal example:
---
title: A template for the *arxiv* style
authors:
  - name: David S. Hippocampus
    thanks: Use footnote for providing further information about author (webpage, alternative address)---*not* for acknowledging funding agencies. Optional.
    department: Department of Computer Science
    affiliation: Cranberry-Lemon University
    location: Pittsburgh, PA 15213
    email: hippo@cs.cranberry-lemon.edu
  - name: Elias D. Striatum
    department: Department of Electrical Engineering
    affiliation: Mount-Sheikh University
    location: Santa Narimana, Levand
    email: stariate@ee.mount-sheikh.edu
abstract: |
  Enter the text of your abstract here.
keywords:
  - blah
  - blee
  - bloo
  - these are optional and can be removed
bibliography: references.bib
output: rticles::arxiv_article
---

# Introduction

\lipsum[2]
\lipsum[3]
```{r, echo=FALSE}
plot(1:10)
```

# Headings: first level
\label{sec:headings}

\lipsum[4] See Section \ref{sec:headings}.

## Headings: second level
\lipsum[5]

\begin{equation}
\xi _{ij}(t)=P(x_{t}=i,x_{t+1}=j|y,v,w;\theta)= {\frac {\alpha _{i}(t)a^{w_t}_{ij}\beta _{j}(t+1)b^{v_{t+1}}_{j}(y_{t+1})}{\sum _{i=1}^{N} \sum _{j=1}^{N} \alpha _{i}(t)a^{w_t}_{ij}\beta _{j}(t+1)b^{v_{t+1}}_{j}(y_{t+1})}}
\end{equation}

### Headings: third level
\lipsum[6]

\paragraph{Paragraph}
\lipsum[7]

# Examples of citations, figures, tables, references
\label{sec:others}

\lipsum[8] some text [@kour2014real; @kour2014fast] and see @hadash2018estimate.

The documentation for \verb+natbib+ may be found at
\begin{center}
  \url{http://mirrors.ctan.org/macros/latex/contrib/natbib/natnotes.pdf}
\end{center}
Of note is the command \verb+\citet+, which produces citations
appropriate for use in inline text.  For example,

\begin{verbatim}
   \citet{hasselmo} investigated\dots
[...]

PR #341, opened 2020-10-28
fix arxiv template issues

Athanasiamo:
This is just a small fix to enable figures from code chunk to be included without error in the arxiv templates (#332 ). 
The skeleton might also need some adapting to make it more Rmarkdown and less latex, but I did not do that, in case the format is intentionally so.


cderv:
Looks good ! Thank you ! 

Could you add a new item to NEWS also ? 

Thank you !

Athanasiamo:
should I add it under the current version number?

cderv:
> should I add it under the current version number?

Yes 0.17 is currently the dev version

(I am just adding [magic keyword](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) so that it will close #332 when we merged this PR)

cderv:
Thank you @Athanasiamo !

inst/rmarkdown/templates/arxiv/skeleton/skeleton.Rmd
```diff
@@ -20,7 +20,9 @@ keywords:
   - bloo
   - these are optional and can be removed
 bibliography: references.bib
-output: rticles::arxiv_article
+output: 
+  rticles::arxiv_article:
+    keep_tex: true
```
cderv, line 25: @Athanasiamo just a precision about this change I kept: 

`keep_tex = TRUE` is already the default in `arxiv_article()`. So even if you don't mention it in the YAML header, the tex file will be kept. 

FYI

Issue #345, opened 2020-11-06
header-includes should be moved outside individual templates


slemonide:
I think it would be reasonable to move this chunk of code outside individual templates:
```
$for(header-includes)$
$header-includes$
$endfor$
```

It should be possible to include additional packages in for any template. Also, it is easy to miss this, as is the case for arxiv template (which has this chunk missing).

cderv:
> move this chunk of code outside individual templates:

What do you think of ? 

These are pandoc templates and they can't inherit from each others. (at least not with older pandoc version). I think we need to have them separate even with common element. 

We could obviously create a generator of Pandoc templates base on some intial template to fill in (to become pandoc template) but this is some work. 

> It should be possible to include additional packages for any template. Also, it is easy to miss this, as is the case for arxiv template (which has this chunk missing).

We missed that when reviewing the Axriv template a while back now. Pleaso open a PR to add those line.

eliocamp:
Perhaps part of the testing process could be that each template has that chunk of code.

cderv:
That is a great idea ! 

I think we would need to have a documentation (maybe a vignette) explaining how to contribute a new format in **rticles**, including what to think about while tranforming the journal tex template to a pandoc template for the package. This would detail which pieces includes for Pandoc to work, and which other common piece are useful to have in each template. 

This relates to https://github.com/rstudio/rticles/pull/346#discussion_r518728148

What do you think ?

cderv:
> Perhaps part of the testing process could be that each template has that chunk of code.

#453 brings tests to check that this is included in all templates. 

It does not for now put it in a partial, but this could be considered if we upgrade the min Pandoc requirement

Issue #257, opened 2019-11-21
How can I use preamble.tex with R Journal template? 


agila5:
---
By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('rticles')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/rticles')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

---

Hi. I have a problem with the `preamble.tex` file used in the Rmarkdown `rticles `template for R Journal Submission. This is the procedure I'm following: 

1. Create a new Rmarkdown `rticles `template for submission to the R Journal. 
1. Knit the document and everything seems to be fine. 
1. Add some latex code to `preamble.tex` (like `\usepackage{subfig}`) 
1. Knit the document and it doesn't compile anymore. 

![immagine](https://user-images.githubusercontent.com/22221146/69355872-d6f6c600-0c82-11ea-9a88-69c46c012074.png)

If I just remove `\usepackage{subfig}` from `preamble.tex` then everything works fine. I think the problem is related to the fact that the `\usepackage{...}` code isn't put in the preamble of the .tex document but I don't know how to fix that. 

If you want to test I created I git repo [here](https://github.com/agila5/test_rticles_RJournal) with all the Rmarkdown files up to point 3 of the previous list. If I knit the Rmd document that I get an error. 

I've already asked this question on [SO](https://stackoverflow.com/questions/58843383/how-to-use-preamble-tex-with-rmarkdown-rticles-template-for-r-journal-submission) and [rstudio community](https://community.rstudio.com/t/how-to-use-preamble-tex-with-rmarkdown-rticles-template-for-r-journal-submission/44824). 

``` r
xfun::session_info("rticles")
#> R version 3.6.1 (2019-07-05)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 18362)
#> 
#> Locale:
[...]

PR #346, opened 2020-11-06
Include header-includes and preamble for arxiv template; add some ins…

slemonide:
…tructions on how to install this package from a local folder (for development/debug/etc)

See issue #345

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,11 @@ If you wish to install the development version from GitHub (which often contains
 remotes::install_github("rstudio/rticles")
 ```
 
+If you want to install a locally modified version, clone this repo and do:
+```r
+remotes::install_local("path/to/local/repo")
+```
+
 ## Overview
 
 The **rticles** package provides a suite of custom [R Markdown](https://rmarkdown.rstudio.com) LaTeX formats and templates for various formats. Most of the templates are provided and maintained by the community, and anyone can contribute a new template. See [How to contribute](#how-to-contribute) below.
--- a/inst/rmarkdown/templates/arxiv/resources/template.tex
+++ b/inst/rmarkdown/templates/arxiv/resources/template.tex
@@ -66,6 +66,13 @@ $if(csl-refs)$
 \newcommand{\CSLIndent}[1]{\hspace{\cslhangindent}#1}
 $endif$
 
+% Pandoc header
+$for(header-includes)$
+$header-includes$
+$endfor$
+
+$preamble$
+
 \begin{document}
 \maketitle
 


README.md
```diff
@@ -15,6 +15,11 @@ If you wish to install the development version from GitHub (which often contains
 remotes::install_github("rstudio/rticles")
 ```
 
+If you want to install a locally modified version, clone this repo and do:
+```r
+remotes::install_local("path/to/local/repo")
+```
+
```
cderv, line 22: This is not specific to **rticles** - this is development practice. You can also use `install.packages` or RStudio IDE included  build & install feature. 

I don't think we should add that here. Why do you suggest that ?

inst/rmarkdown/templates/arxiv/resources/template.tex
```diff
@@ -66,6 +66,13 @@
 \newcommand{\CSLIndent}[1]{\hspace{\cslhangindent}#1}
 $endif$
 
+% Pandoc header
+$for(header-includes)$
+$header-includes$
+$endfor$
+
+$preamble$
+
```
cderv, line 75: Why add a preamble variable ? 
`header-includes` variable is used for that already. Using `includes` argument in the yaml header for example.
https://bookdown.org/yihui/rmarkdown/pdf-document.html#includes-1

cderv (changes requested):

README.md
slemonide: I think it's good to leave a note on how to modify and install a local development version. This would make it easier for people to contribute. Maybe move this under "Development" or "Contribute" section?

inst/rmarkdown/templates/arxiv/resources/template.tex
slemonide: This makes it consistent with other templates.

Also people expect for `preamble` to work: https://github.com/rstudio/rticles/issues/257#issuecomment-564304252

README.md
cderv: It we keep something like that it should be moved to contribute section yes. I would prefer to add a link to some documentation about how to work and develop r package like https://r-pkgs.org/index.html instead of developing  a new part on how to develop.

inst/rmarkdown/templates/arxiv/resources/template.tex
cderv: Oh I see. so this is a specificity of some **rticles** templates. Not all have it, and it won't work with **rmarkdown**, so it could be expected but by some **rticles** user only. 

Fine by me. Thank you for pointing this out!

README.md
yihui: I'd also prefer leaving this out. As @cderv said, this is development practice and not specific to the **rticles** package. There are several ways to install a local package, and `remotes::install_local()` is one of them. RStudio users will probably find `Ctrl/Cmd + Shift + B` easier.

yihui (changes requested):
I think this PR is good to merge after documenting the `preamble` option and leaving out the changes in README.md (if you feel it is really necessary, please feel free to submit a separate PR, so it won't block this PR from being merged). Thanks!

inst/rmarkdown/templates/arxiv/resources/template.tex
```diff
@@ -66,6 +66,13 @@
 \newcommand{\CSLIndent}[1]{\hspace{\cslhangindent}#1}
 $endif$
 
+% Pandoc header
+$for(header-includes)$
+$header-includes$
+$endfor$
+
+$preamble$
```
cderv, line 74: `preamble` and `header-includes` are the same IMO. I don't think a template should have both - some of existing template have both but not sure why. 

Why do you think we need both ? 

I would rather only add `header-includes` (which #356 also suggest).

cderv:
The `header-include` variable has been added to the template already in https://github.com/rstudio/rticles/pull/356 

I'll close this PR then.

PR #348, opened 2020-11-09
Updates AMS to V5

eliocamp:
Following #298.
I wanted to push this only after submitting and being given the A-OK from the good people at the AMS. But now it turns out I won't be submitting to that journal, so, well, here's the updated version of the AMS article template.


cderv:
Thanks @eliocamp ! I see there is hook you mentioned in #298. Is this a modified version of `knitr::hook_plot_tex()` ? 

@yufree, as the original contributor, do you want to take a look at this too ?

eliocamp:
> Thanks @eliocamp ! I see there is hook you mentioned in #298. Is this a modified version of `knitr::hook_plot_tex()` ?

Yep. The change is the that it uses appendcaption instead of caption. Is part of the instructions from the AMS template.

yufree:
It seems the check could not pass: 
```
! Package inputenc Error: Unicode character ≤ (U+2264)

  (inputenc)                not set up for use with LaTeX.
```
To solve this issue, you could change the latex engine from `pdflatex` to `xelatex` in the template to avoid the issue.

inst/rmarkdown/templates/ams/skeleton/skeleton.Rmd
```diff
@@ -1,62 +1,72 @@
...
-bibliography: amstest.bib
-csl: american-meteorological-society.csl
+  Enter the text of your abstract here. This is a sample American Meteorological Society (AMS) \LaTeX\ template.  This document provides authors with instructions on the use of the AMS \LaTeX\ template.  Authors should refer to the file amspaper.tex to review the actual \LaTeX\ code used to create this document. The template.tex file should be modified by authors for their own manuscript.
+statement: |
+  An optional plain language summary suitable for wider readership that provides additional context for why the work is relevant to science and society. It should be concise (≤120 words) and authors should follow the guidelines for significance statements (\url{https://www.ametsoc.org/index.cfm/ams/publications/authors/journal-and-bams-authors/significance-statements/}) (currently MWR,WAF, and WCAS only).
+capsule: |
+   A short (20-30 words maximum) capsule summary of your article (for BAMS only).
+bibliography: references.bib
 output: rticles::ams_article
```
yufree, line 21: change here to pass the check.

```
output: 
  rticles::ams_article:
    latex_engine: xelatex
```

cderv: Does it means this format has to default to xelatex ?

yufree: Just for this template. It seems the bib file contain unicode character which will not map to output through pdflatex. Such issue blocked the check. Since the bib file might be from ASM, we could set the default engine to xelatex to be friendly to the user.

eliocamp:
I added xelatex as latex_engine in the skeleton.

yihui:
> Yep. The change is that it uses appendcaption instead of caption. Is part of the instructions from the AMS template.

@eliocamp If you had to copy `hook_plot_tex` from **knitr** and modify it in this package, I'll definitely reconsider https://github.com/yihui/knitr/issues/1872, because this will be a disaster for us to maintain in **rticles** :)

cderv:
Sorry for the delay. V6 template for AMS is out. Should we consider V6 and abandon V5 here ? 
https://github.com/rstudio/rticles/issues/298#issuecomment-914373266

cderv:
I am closing this PR in favor of #444. 

Thanks.

PR #349, opened 2020-11-12
use gha instead of travis

cderv:
Customizing actions from r-lib/actions one. 

Unfortunately, not sure we can cache CTAN packages. One way to deal with that would be to use a set of docker images with with pre-installed component. 

We'll how long it takes with this config


cderv:
@yihui just a question about `ctex` and Chinese fonts. 

Currently it errors on Windows and Mac because fonts are not found: 
* Mac - Kaiti SC: https://github.com/rstudio/rticles/runs/1391411927?check_suite_focus=true#step:14:139
* Windows - SimHei: https://github.com/rstudio/rticles/runs/1391411968?check_suite_focus=true#step:14:221

Do you know if these fonts can be added easily on those OS by command line ? 
If not, don't bother looking - I'll just skip this test of ctex format on Windows and Mac to just run it on Linux. 

Thank you.

cderv:
I am merging this, but if you have the trick for the fonts, feel free to tell me and I'll add it. 

Thanks

yihui:
I know how to make it work on macOS and Windows (we could use https://ctan.org/pkg/fandol), but I think it's okay for the ctex test to be run only on Linux. Thanks!

cderv:
I leave it that way then

PR #356, opened 2020-12-15
Adds some more variables to arxive template

eliocamp:
Related to #345
The arxive template didn't have the holy 

```
$for(header-includes)$
$header-includes$
$endfor$
```

it also had biblyographystyle hardcoded.

PR #358, opened 2020-12-16
Uses biblio-style instead of bibstyle in springer_article

eliocamp:
Following on this comment
https://github.com/rstudio/rticles/pull/356#discussion_r544181296

This changes the variable name from bibstyle to biblio-style to match the default pandoc template for latex and other rticles templates.

### DIFF
--- a/inst/rmarkdown/templates/springer/resources/template.tex
+++ b/inst/rmarkdown/templates/springer/resources/template.tex
@@ -132,7 +132,8 @@ $endif$
 
 $body$
 
-\bibliographystyle{$if(biblio-style)$$biblio-style$$else$spbasic$endif$}
+
+\bibliographystyle{$if(biblio-style)$$biblio-style$$elseif(bibstyle)$$bibstyle$$else$spbasic$endif$}
 \bibliography{$for(bibliography)$$bibliography$$sep$,$endfor$}
 
 \end{document}
--- a/inst/rmarkdown/templates/springer/resources/template.tex
+++ b/inst/rmarkdown/templates/springer/resources/template.tex
@@ -132,7 +132,7 @@ $endif$
 
 $body$
 
-\bibliographystyle{$if(bibstyle)$$bibstyle$$else$spbasic$endif$}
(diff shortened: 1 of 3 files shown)


cderv:
Thanks @eliocamp ! 

No so easy to do because you don't want to break every existing document that are using `bibstyle`. At least, I don't want to 😄 

I believe this could be done if both are supported - at least for some versions as a deprecated function or argument.

eliocamp:
Yes, I see. That's one reason I think it might be useful to copy the template.tex file in the same folder as the Rmd file. That way, each rticle project is setf-contained and can't break (easily) with changes in rticles.

I pushed some more subtle logic to catch the old behaviour. Would that work?

inst/rmarkdown/templates/springer/resources/template.tex
```diff
@@ -132,7 +132,8 @@
 
 $body$
 
-\bibliographystyle{$if(biblio-style)$$biblio-style$$else$spbasic$endif$}
+
+\bibliographystyle{$if(biblio-style)$$biblio-style$$elseif(bibstyle)$$bibstyle$$else$spbasic$endif$}
```
cderv, line 136: Yes it would but unfortunately for us `elsif` is only supported since Pandoc 2.8
https://pandoc.org/releases.html#pandoc-2.8-2019-11-22

So we need to use good old nested if else structure.

eliocamp: image

cderv:
> That's one reason I think it might be useful to copy the template.tex file in the same folder as the Rmd file. That way, each rticle project is setf-contained and can't break (easily) with changes in rticles.

But you would not get the updates in templates that are required when TeX Live updates, or Pandoc updates, etc... 
having the template in the package also allow to hide some of the complexity behind all this. Like in Pandoc itself with the default template, or in **rmarkdown** or **bookdown** where you use the included templates that live in the packages.


However, advanced users are free to use their templates. So being able to opt-in to self contained mode when creating a template could be useful maybe. You can open a specific issue on that if you want.

PR #372, opened 2021-02-04
add `ims_article()`

auzaheta:

- [X] Unless you have done it in any other RStudio's projects before, please sign the [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). You can send the signed copy to <jj@rstudio.com>.

- [X] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [X] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname/resources/template.tex`.

- [X] Add a skeleton article `inst/rmarkdown/templates/journalname/skeleton/skeleton.Rmd`.

- [X] Add a description of the template `inst/rmarkdown/templates/journalname/template.yaml`.

- [X] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. If you are using TinyTeX or TeX Live, you can verify if a package is available on CTAN via `tinytex::parse_packages(files = "FILENAME"")` (e.g., when `FILENAME` is `plain.bst`, it should return `"bibtex"`, which means this file is from a standard CTAN package). Please keep the number of new files absolutely minimal (e.g., do not include PDF output files), and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include too many items in it without using all of them).

- [X] Update NEWS.

- [X] Update README with a link to the newly supported journal. Please add your Github username and the full name of the journal (follow other examples in the list). 

- [X] Add a test to `tests/testit/test-formats.R` by adding a line `test_format("journalname")`. We try to keep them in alphabetical order.

Thank you!


inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+                     & = & \mu_{\text{normal}} - Y_{x/s}\bigl(1-H(C_{s})\bigr)(m_{s}+\pi /Y_{p/s})\\
+                     & = & \mu_{\text{normal}}/Y_{x/s}+ H(C_{s}) (m_{s}+ \pi /Y_{p/s}).\nonumber
+\end{eqnarray}
+
+<!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+<!-- %% Example with single Appendix:            %% -->
+<!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+
+\begin{appendix}
+\section*{Title}\label{appn} %% if no title is needed, leave empty \section*{}.
```
cderv, line 344: You can use markdown from that too
```suggestion
# Title {.unnumbered #appn}
```
* #id to use as label
* .unnumbered so that the `*` is added to `\section`

And the following content can also use markdown for all the section and labels.

cderv (changes requested):
Thanks for the contribution ! 

First of all, I understand that the article is based on the template of AOAS journal. Do you think or know if this is generic to every journals they have ? I see several: https://www.e-publications.org/ims/support/ims-instructions.html and did not compare the templatex `.tex`

Does changing only the `journal: aoas` field in the YAML will work for any journal they have ? 

I ask because usually the function is `journalname_article()`. So it seems it should be called `aosas_article` instead as this is the name of the journal. But if what I said before, maybe we could make an exception. 

If not we could keep one function, but may need to modify a bit. Like providing an argument to the function `ims_article(journal = "aoas)` that would switch between the correct template for IMS and using the share component. If there is no shared component, then maybe it is one function in **rticles** by journal.

For the rest of the comments, see below. There is a few things that I think needs to change (depending on your feedback). For the rest this is mainly for discussion before merging.

For example ,I made some suggestion to include a bit less LaTeX syntax and use more markdown in the skeleton. Pandoc and **rmarkdown** allow to write some markdown syntax to generate the expected latex. What do you think ?

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+This is the second paragraph of the quotation.  It is just
+as dull as the first paragraph.
+\end{quotation}
+
+# Environments
+
+\subsection{Examples for \emph{\texttt{plain}}-style environments}
+\begin{axiom}\label{ax1}
+This is the body of Axiom \ref{ax1}.
+\end{axiom}
```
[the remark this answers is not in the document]
auzaheta: Thanks for the tip, I didn't know this Rmarkdown syntax. However, I had a problem. When I try to compile the pdf from the `skeleton.rmd`, it doesn't recognize this syntax. On the other hand, if I use `bookdown::pdf_book` using the `template.tex` it does the trick. The same problem for `# Title {.unnumbered}`.

I'm wondering if I should add something to the `ims_article()` or maybe I'm missing something here.

An additional question: How can I translate to this syntax this chunk of code:

```latex
\begin{proof}[Proof of Lemma \ref{le1}]
This is the body of the proof of Lemma \ref{le1}.
\end{proof}
```

and this title section

```latex
\subsection{First subsection of Appendix \protect\ref{appB}}
```

I updated Rmardown to the one in GitHub. Some additional info, just in case:

```r
sessioninfo::session_info()
- Session info --------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 4.0.3 (2020-10-10)
 os       Windows 10 x64              
 system   x86_64, mingw32             
 ui       RStudio                     
 language (EN)                        
 collate  German_Switzerland.1252     
 ctype    German_Switzerland.1252     
 tz       Europe/Berlin               
 date     2021-02-10                  

- Packages ------------------------------------------------------------------------------------------------------------------
 ! package     * version date       lib source                            
   assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.0.2)                    
   callr         3.5.1   2020-10-13 [1] CRAN (R 4.0.3)                    
   cli           2.3.0   2021-01-31 [1] CRAN (R 4.0.3)                    
   crayon        1.4.1   2021-02-08 [1] CRAN (R 4.0.3)                    
   curl          4.3     2019-12-02 [1] CRAN (R 4.0.2)                    
   digest        0.6.27  2020-10-24 [1] CRAN (R 4.0.3)                    
   evaluate      0.14    2019-05-28 [1] CRAN (R 4.0.2)                    
   fortunes      1.5-4   2016-12-29 [1] CRAN (R 4.0.0)                    
   glue          1.4.2   2020-08-27 [1] CRAN (R 4.0.2)                    
   htmltools     0.5.1.1 2021-01-22 [1] CRAN (R 4.0.3)                    
   knitr         1.31    2021-01-27 [1] CRAN (R 4.0.3)                    
   pkgbuild      1.2.0   2020-12-15 [1] CRAN (R 4.0.3)                    
   prettyunits   1.1.1   2020-01-24 [1] CRAN (R 4.0.2)                    
   processx      3.4.5   2020-11-30 [1] CRAN (R 4.0.3)                    
   ps            1.5.0   2020-12-05 [1] CRAN (R 4.0.3)                    
   R6            2.5.0   2020-10-28 [1] CRAN (R 4.0.3)                    
   remotes       2.2.0   2020-07-21 [1] CRAN (R 4.0.2)                    
   rlang         0.4.10  2020-12-30 [1] CRAN (R 4.0.3)                    
 V rmarkdown     2.6     2021-02-10 [1] Github (rstudio/rmarkdown@a62cb20)
   rprojroot     2.0.2   2020-11-15 [1] CRAN (R 4.0.3)                    
   rstudioapi    0.13    2020-11-12 [1] CRAN (R 4.0.3)                    
   rticles     * 0.18.3  2021-02-10 [1] local                             
   sessioninfo   1.1.1   2018-11-05 [1] CRAN (R 4.0.2)                    
   withr         2.4.1   2021-01-26 [1] CRAN (R 4.0.3)                    
   xfun          0.20    2021-01-06 [1] CRAN (R 4.0.3)                    
   yaml          2.2.1   2020-02-01 [1] CRAN (R 4.0.2)                    

[1] C:/Users/ualvaro/AppData/Local/Programs/R-4.0.3/library

 V -- Loaded and on-disk version mismatch.
```

inst/rmarkdown/templates/ims/resources/template.tex
```diff
@@ -0,0 +1,196 @@
...
+
+
+\startlocaldefs
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                          %%
+% Uncomment next line to change            %%
+% the type of equation numbering           %%
+%                                          %%
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\numberwithin{equation}{section}
```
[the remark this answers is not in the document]
auzaheta: I [added](https://github.com/auzaheta/rticles/blob/bc5f41196104b1c398134af064be6af7141f215c/inst/rmarkdown/templates/ims/resources/template.tex#L113) `predefined-theoremstyle` variable to activate/deactive this chunk.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+  your own.
+keywords: 
+  - First keyword
+  - Second keyword
+  
+bibliography: ims.bib
+biblio-style: imsart-nameyear # imsart-number
+citation_package: natbib
+output: rticles::ims_article
+toc: false # Please use for articles with 50 pages and more
```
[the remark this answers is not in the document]
auzaheta: Thanks for the suggestion, I followed your suggestion.

inst/rmarkdown/templates/ims/skeleton/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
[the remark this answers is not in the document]
auzaheta: I moved the file to the parent folder. I wasn't sure about the proper way to include this file.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+  This is a sample input file.  Comparing it with the output it
+  generates can show you how to produce a simple document of
+  your own.
+keywords: 
+  - First keyword
+  - Second keyword
+  
+bibliography: ims.bib
+biblio-style: imsart-nameyear # imsart-number
+citation_package: natbib
```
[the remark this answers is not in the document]
auzaheta: The template is supposed to work only with `natbib`. In accordance, I removed the `citation_package` field in the YAML. 

On the other hand, I couldn't make that Rmarkdown generates the cited references with a hyperlink to the references section (as in the journal example).  Also, as far as I know, the Rmarkdown parsing would generate the citations using `\citep` and similars, oppose to the `\citeauthoryear` used in the tex template by the journal. I'm afraid this could be a problem with the journal. 

For now, I'm using in the skeleton the commands recommended in the journal template. I also added the content for Pandoc citeproc in the template, but It's commented bc the mention reasons above. Everything would work uncommenting those lines, but it won't produce the expected output. 

I'm happy to read any suggestion about, maybe from other template experiences.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+\item[] $\mathsf{ABCabc123}$
+\item[] $\mathtt{ABCabc123}$
+\item[] $\mathfrak{ABCabc123}$
+\end{itemize}
+Note that \verb|\mathcal, \mathbb| belongs to capital letters-only font typefaces.
+
+# Notes
+
+Footnotes[^1]
+pose no problem.\footnote{Note that footnote number is after punctuation.}
```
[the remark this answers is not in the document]
auzaheta: I changed it to Rmarkdown too.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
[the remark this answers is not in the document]
auzaheta: Thanks a lot for the tip. I didn't know this syntax. I changed most of them to this new syntax. I also added a comment pointing out that this syntax only works with `Rmarkdown >= 2.7`, I hope this is not a problem for those using the CRAN version.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+          & $\beta_1$   & 0.79   & 0.21 & 0.38   & 0.78   & 1.20  \\
+          & $\beta_2$   & $-$0.28  & 0.10 & $-$0.48  & $-$0.28  & $-$0.07 \\[6pt]
+{Model 2} & $\beta_0$   & $-$11.85 & 2.24 & $-$17.34 & $-$11.60 & $-$7.85 \\
+          & $\beta_1$   & 0.73   & 0.21 & 0.32   & 0.73   & 1.16  \\
+          & $\beta_2$   & $-$0.60  & 0.14 & $-$0.88  & $-$0.60  & $-$0.34 \\
+          & $\beta_3$   & 0.22   & 0.17 & $-$0.10  & 0.22   & 0.55  \\
+\hline
+\end{tabular}
+%
+\end{table}
```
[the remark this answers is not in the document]
auzaheta: Thanks for the suggestion, I implemented it.

auzaheta:
Thanks a lot for the suggestions to improve the template and the detailed review.

The template is generic for all the journals they have; being precise, all templates use the same tex package to create the articles templates. I checked the five journals directly related to [IMS Journals and publications](https://www.e-publications.org/ims/support/ims-instructions.html), and I added a comment in the skeleton indicating the acronym of the journals and additional [variable](https://github.com/auzaheta/rticles/blob/bc5f41196104b1c398134af064be6af7141f215c/inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd#L46) that should be used for some of them.  I didn't find a way of adding an example of this variable's content that allows an easy uncommenting way.

In summary, it works for these five journals. All of them have the same tex template. I tried to emphasize the *annals of applied statistics* in the news and the readme because I think it is the one most of us could use.

This template could be used for other journals beyond these five (also IMS). It would require developing the function with the argument with an indication of the journal and create independent skeletons (as you suggested). However, this is beyond what I know how to do and what I will need soon. Happy to learn how to do it in any case.

I tried to follow your recommendation to have more Rmarkdown syntax in the skeleton.

inst/rmarkdown/templates/ims/resources/template.tex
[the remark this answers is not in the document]
cderv: Nice! Should the `\startlocaldefs` and `\endlocaldefs` be included inside the if ? 
Because if it is false, it would lead to this in the tex file : 
```tex
\startlocaldefs
\endlocaldefs
```
that would not be useful right ?

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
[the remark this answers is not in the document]
cderv: If this template is supposed to work only with `natbib` then, the function `ims_article` should have that by default for `citation_package`

```r
ims_article <- function(..., keep_tex = TRUE, citation_package = "natbib") {
  pdf_document_format(
    "ims", keep_tex = keep_tex, citation_package = citation_package, ...
  )
}
```

Does it makes sense ? Otherwise, natbib won't be used by Pandoc. Currently, it is Pandoc citeproc which is used, not natbib directly.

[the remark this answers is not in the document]
auzaheta: Thanks for the clarification. I changed the function and template accordingly.

inst/rmarkdown/templates/ims/resources/template.tex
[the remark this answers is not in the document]
auzaheta: That's right, I had made the change

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,422 @@
...
+and longer ones.
+
+<!-- The following syntax works with Rmarkdown >= 2.7 -->
+<!-- ::: {.environment latex=true} -->
+<!--   code -->
+<!-- ::: -->
+<!-- It's equivalent to  -->
+<!-- \begin{environment} -->
+<!--   % code -->
+<!-- \end{environment} -->
```
cderv, line 201: You are right. I went a bit ahead of what is available. 
The current syntax that works is : 
````markdown
::: {.environment data-latex=""}
   code
:::
````
Doc: https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html

We added a simplified version in dev **rmarkdown**. We may use the current working version instead and change when 2.7 in on CRAN. What do you think ? 

Thanks for adding this note in the skeleton!

File: inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd [hunk 10]
```diff
@@ -0,0 +1,422 @@
...
+
+  This is a sample input file.  Comparing it with the output it
+  generates can show you how to produce a simple document of
+  your own.
+keyword-subClass: |
+  <!-- \begin{keyword}[class=MSC2020] % Uncomment for aap, aop, aos journals -->
+  <!-- \kwd[Primary ]{00X00} -->
+  <!-- \kwd{00X00} -->
+  <!-- \kwd[; secondary ]{00X00} -->
+  <!-- \end{keyword} -->
```
cderv, line 50: If I understand correctly, this part is only for 3 of the five journal ? 

I think it is fine as you set it but maybe the key should also be commented not the content. Like 
```suggestion
# Uncomment below for aap, aop, aos journals
# keyword-subClass: |
#  \begin{keyword}[class=MSC2020]
#  \kwd[Primary ]{00X00}
#  \kwd{00X00}
#  \kwd[; secondary ]{00X00}
#  \end{keyword}
```

Also, maybe keep everything lowercase for consistency ? `keyword-subclass` ? 

The way to do it without commenting / commenting would be to have the journal being a variable in the function 
````r
output:
  ims_article:
    journal: aap
````
You could then check the journal name and set
* the journal variable for the template as it is now
* Set a `keyword-subclass` variable that would activate this part in the template. 

It is not necessary to do this change, but just sharing how I would do it.

**Review by cderv — COMMENTED** [review]
> In summary, it works for these five journals. All of them have the same tex template. I tried to emphasize the annals of applied statistics in the news and the readme because I think it is the one most of us could use.

That is really great news ! We can keep `ims_article()` then and have the journal variable for switching between journal. Maybe we should add somewhere that it _should_ work for the five journals and list them ? maybe add the link of  https://www.e-publications.org/ims/support/ims-instructions.html ? 
It could be in the function help page.

> I didn't find a way of adding an example of this variable's type of content that allows an easy way of uncommenting it.

See below

> I tried to follow your recommendation to have more Rmarkdown syntax in the skeleton.

Awesome! Thank you. 

This is very cool. Thanks for this contribution! 

Only last thing is indeed the **rmarkdown** 2.7 dependency. I let you see the commend below and let you time to react before merging.

File: inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd (hunk 9 above)
Comment (auzaheta) [reply inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd]: Thanks for the hint. Now it uses the current working version

File: inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd (hunk 10 above)
Comment (auzaheta) [reply inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd]: Thanks for the tip. Now the `ims_article()` includes the new parameter and sets a new variable that activates this part in the template. Also, the template variable changed to `keyword-subclass`

File: inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd (hunk 9 above)
Comment (auzaheta) [reply inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd]: It's updated to work with the current syntax. Thanks for pointing out the documentation.

File: R/article.R [hunk 11]
```diff
@@ -175,14 +175,48 @@ frontiers_article <- function(..., keep_tex = TRUE) {
...
+
+  args <- c(
+    "journal" = journal,
+    if (with_kwsc) c("with_kwsc" = with_kwsc)
+    )
+
+  # pandoc_variable_arg not exported from rmarkdown, adapted from ieee_article()
+  pandoc_arg_variable <- function(var_name, value) {
+    c("-V", paste0(var_name, "=", value))
+  }
```
cderv, line 212: The exported function is `rmarkdown::pandoc_variable_arg()`.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,414 @@
...
+
+  This is a sample input file.  Comparing it with the output it
+  generates can show you how to produce a simple document of
+  your own.
+keyword-subclass: | 
+ \begin{keyword}[class=MSC2020] % It must be define for aap, aop, aos journals. For aoas, sts is not used
+ \kwd[Primary ]{00X00}
+ \kwd{00X00}
+ \kwd[; secondary ]{00X00}
+ \end{keyword}
```
cderv, line 49: Is this fixed for the templates ? or should it be modified by the user ? 
I was under the impression of the former and then it should be in the `template.tex` but I may be wrong ?

R/article.R
auzaheta: Thanks for pointing it out. I took it from the `ieee_article` function

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
auzaheta: It should be modified by the user.

cderv: Oh ok. So it made sense to have it commented or not commented.

I saw you followed the idea I suggested to have the function handle the variable. I suggested that thinking it would be removed from the YAML. 😅 

Anyway, if you're happy with the change to provide the `journal` in the function, it is ok with me too.

R/article.R
cderv: Thanks ! I'll make the change there.

inst/rmarkdown/templates/ims/skeleton/skeleton.Rmd
auzaheta: Thanks for the help, I'm happy with the change, I hope it makes it easier for users

cderv (approved):
Thanks a lot!

Issue #383, opened 2021-04-12
Issue with Pandoc CSL refs and some articles templates


cderv:
Reported first in tinytex https://github.com/yihui/tinytex/issues/298

````r
> dir.create(tmp_dir <- tempfile())
> owd <- setwd(tmp_dir)
> rmarkdown::draft("myArticle", template = "frontiers", package = "rticles", edit = FALSE)
> list.files("myArticle/")
> setwd("myArticle/")
> rmarkdown::render("myArticle.Rmd", quiet = TRUE)
! You can't use `\vadjust' in vertical mode.
l.300 \leavevmode\vadjust
                          pre{\hypertarget{ref-Neurobot2013}{}}% 

Error: LaTeX failed to compile myArticle.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips. See myArticle.log for more info.
> setwd(owd)
> unlink(tmp_dir, recursive = TRUE)
````

Error seems to be from Pandocs CSL references part
````
\begin{CSLReferences}{1}{0}
\leavevmode\vadjust pre{\hypertarget{ref-Neurobot2013}{}}%
Author4, N., and Author5, N. (2013). Title of the article.
\emph{Frontiers in Neurorobotics} 7.
doi:\href{https://doi.org/10.3389/fnbot.2013.56789}{10.3389/fnbot.2013.56789}.

\leavevmode\vadjust pre{\hypertarget{ref-Neuro2013}{}}%
LastName1, A., LastName2, A., and LastName2, A. (2013). Article title.
\emph{Frontiers in Neuroscience} 30, 10127--10134.
doi:\href{https://doi.org/10.3389/fnins.2013.12345}{10.3389/fnins.2013.12345}.

\leavevmode\vadjust pre{\hypertarget{ref-Gene2012}{}}%
OtherAuthor, N., and Coauthor, N. S. (2012). Article title.
\emph{Frontiers in Genetics} 30, 16417--16418.
doi:\href{https://doi.org/10.3389/fgene.2012.54321}{10.3389/fgene.2012.54321}.

\end{CSLReferences}
````

Same with `elsevier` template
```r
rmarkdown::draft("myArticle", template = "elsevier", package = "rticles", edit = FALSE)
```

cderv:
It seems new from Pandoc 2.13 as with previous version is work ok. (tested locally with 2.13 then 2.11.4)

I added a test for Pandoc devel to confirm:

[...]

PR #390, opened 2021-04-22
Update CSLreferences usage in line with upstream

BlackEdder:
I had a problem with my rmarkdown not compiling to pdf any more using the newest pandoc (see here for more details: https://github.com/jgm/pandoc/issues/7215). It turned out that this was due to outdate csl-refs in the elsevier template. 

The first commit in this pull request fixes this in the elsevier template and this was confirmed to solve the problem for me. The second commit also fixes it in all other template files, but I have been unable to test this. The fix is very straightforward though, so I don't foresee any issues.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rticles
 Type: Package
 Title: Article Formats for R Markdown
-Version: 0.19.1
+Version: 0.19.2
 Authors@R: c(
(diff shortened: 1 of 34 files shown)


cderv:
Thank you ! I did not had time to look into it yet, but it seems you've fixed #383 with this PR. 🎉 
Did you check if there was other modification from the template to port here ? 
I just need to check that before merging. 

And please can you sign the CLA if you don't mind ? thank you.

cderv (approved):

PR #391, opened 2021-04-23
Copernicus template update - fixes syntax highlight problem

RLumSK:
Yet another update for the Copernicus Publication template after the `*.tex` produced by 'rticles' bounced back from the journal's office telling that additional LaTeX packages and commands are not supported. This was caused by the default syntax highlight enabled in `rmarkdown::pdf_document()`, which uses additional LaTeX commands.

To ensure that this update aligns with the publisher guidelines, I first discussed with @nuest via email and then contacted the journal's office before making this update (@nuest was in CC). 

Hence, this update ensures that the publisher will accept the `*.tex` produced by the templates without requiring further manual changes by the authors.  

**More details**

* The template now supports the `highlight` parameter of `rmarkdown::pdf_document()`. The default is set to `NULL`, which lets Pandoc set programming code in the verbatim environment. This solution is supported by the publisher, who currently uses the LaTeX packages `algorithms` and `algorithmicx`. The author still has the option to use these two LaTeX packages.
* The `*.Rmd` template and the `*.tex` template were modified accordingly.  
* The options `highlights` and `keep_tex` are now explicitly mentioned in the YAML header

- [x] Update NEWS.

@nuest, I hope I did everything correctly. However, would you mind looking over it another time?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 rticles 0.20
 ---------------------------------------------------------------------
+- Fixes a problem in the Copernicus Publications template that causes a manuscript to bounce back to the author during the typesetting step. This was triggered by the inclusion (unknown to the author) of non-supported LaTeX packages and commands if programming code was part of the manuscript using the standard 'rmarkdown' code support. Additionally, the template gained support for the `highlight` parameter of `rmarkdown::pdf_document` to enable or disable syntax 
+highlight (disabled by default, as required by Copernicus)
 
 
 rticles 0.19
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 rticles 0.20
 ---------------------------------------------------------------------
 - Fixes a problem in the Copernicus Publications template that causes a manuscript to bounce back to the author during the typesetting step. This was triggered by the inclusion (unknown to the author) of non-supported LaTeX packages and commands if programming code was part of the manuscript using the standard 'rmarkdown' code support. Additionally, the template gained support for the `highlight` parameter of `rmarkdown::pdf_document` to enable or disable syntax 
-highlight (disabled by default, as required by Copernicus)
+highlight (disabled by default, as required by Copernicus). **Note: this is breaking change.** In order to see the previous
+syntax highlighting, please set `highlight: "default"` in the YAML header.
 
 
 rticles 0.19
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,12 +1,14 @@
 rticles 0.20
 ---------------------------------------------------------------------
-- Fixes a problem in the Copernicus Publications template that causes a manuscript to bounce back to the author during the typesetting step. This was triggered by the inclusion (unknown to the author) of non-supported LaTeX packages and commands if programming code was part of the manuscript using the standard 'rmarkdown' code support. Additionally, the template gained support for the `highlight` parameter of `rmarkdown::pdf_document` to enable or disable syntax 
-highlight (disabled by default, as required by Copernicus). **Note: this is breaking change.** In order to see the previous
-syntax highlighting, please set `highlight: "default"` in the YAML header.
+
+- Update Copernicus Publications template to comply with editor's guidelines following a manuscript bounce back during the typesetting step. Copernicus does not allow to add any `\usepackage` command as they all are included in `copernicus.cls` for supported LaTeX packages. **This is leading to breaking changes with existing template - please follow the advice below**.
+  - `algorithms: true` cannot be used anymore and as no more effect. `\usepackage{algorithmic}` and `\usepackage{algorithm}` has been removed from the template as the command are already done in `copernicus.cls`. Please, make sure `algorithms` and `algorithmcx` are installed.  
+  - Additionally, the template gained support for the `highlight` parameter of `rmarkdown::pdf_document` to enable or disable syntax highlighting with Pandoc. To comply with the above guideline by Copernicus, it is disabled by default (`highlight: NULL`) to prevent Pandoc adding any more packages required for its highlighting. Syntax highlighting can be reactivating by using `highlight: "default"` in the YAML header as this can be desirable before submitting for typesetting. (thanks, @RLumSK, @nuest, #391).
 
 
 rticles 0.19
 ---------------------------------------------------------------------
+
 - Update Copernicus Publications template to version 6.2 from 2021-01-15 (thanks, @RLumSK, #366).
 
 - Add article template `pihph_article()` for the *Papers in Historical Phonology* (PiHPh) (thanks, @stefanocoretta, #362).
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -57,7 +57,7 @@ copernicus_article <- function(
   )
 ) {
   if("extra_dependencies" %in% names(list(...)))
-    stop("Copernicus does not support additional LaTeX packages and options!
+    warning("Copernicus does not support additional LaTeX packages and options!
           >> Please remove 'extra_dependencies' from your YAML header!",
          call. = FALSE)
 
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -48,14 +48,14 @@
 #' }
 #' @export
 copernicus_article <- function(
-  ..., keep_tex = TRUE, citation_package = "natbib", md_extensions = c(
+  ..., keep_tex = TRUE, highlight = NULL, citation_package = "natbib", md_extensions = c(
     "-autolink_bare_uris", # disables automatic links, needed for plain email in \correspondence
     "-auto_identifiers"    # disables \hypertarget commands
   )
 ) {
   pdf_document_format(
     "copernicus", citation_package = citation_package,
-    keep_tex = keep_tex, md_extensions = md_extensions, ...
+    keep_tex = keep_tex, highlight = highlight, md_extensions = md_extensions, ...
   )
 }
 
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -12,7 +12,7 @@
 #'
 #' An number of required and optional manuscript sections, e.g. \code{acknowledgements}, \code{competinginterests}, or \code{authorcontribution}, must be declared using the respective properties of the R Markdown header - see skeleton file.
 #'
-#' \strong{Version:} Based on Copernicus_package.zip in the version 6.2, 15 January 2021, using \code{copernicus.cls} in version 9.25.
+#' \strong{Version:} Based on \code{copernicus_package.zip} in the version 6.2, 15 January 2021, using \code{copernicus.cls} in version 9.25.
 #'
 #' \strong{Copernicus journal abbreviations:} You can use the function \code{copernicus_journal_abbreviations()} to get the journal abbreviation for all journals supported by the Copernicus article template.
 #'
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -12,13 +12,13 @@
 #'
 #' An number of required and optional manuscript sections, e.g. \code{acknowledgements}, \code{competinginterests}, or \code{authorcontribution}, must be declared using the respective properties of the R Markdown header - see skeleton file.
 #'
-#' \strong{Version:} Based on copernicus_package.zip in the version 5.3, 18 February 2019, using \code{copernicus.cls} in version 8.82.
+#' \strong{Version:} Based on Copernicus_package.zip in the version 6.2, 15 January 2021, using \code{copernicus.cls} in version 9.25.
 #'
-#' \strong{Copernicus journal abbreviations:} You can use the function \code{copernicus_journal_abbreviations()} to get the journal abbreviation for all journals supported by the copernicus article template.
+#' \strong{Copernicus journal abbreviations:} You can use the function \code{copernicus_journal_abbreviations()} to get the journal abbreviation for all journals supported by the Copernicus article template.
 #'
 #' \strong{Important note:} The online guidelines by Copernicus are the official resource.
 #' Copernicus is not responsible for the community contributions made to support the template in this package.
-#' Copenicus converts all typeset TeX files into XML, the expressions and markups have to be highly standardized.
+#' Copernicus converts all typeset TeX files into XML, the expressions and markups have to be highly standardized.
 #' Therefore, please keep the following in mind:
 #'
 #' \itemize{
@@ -26,7 +26,7 @@
 #'   \item Please use only commands in which words, numbers, etc. are within braces (e.g. \code{\\textrm{TEXT}} instead of \code{{\\rm TEXT}}).
 #'   \item For algorithms, please use the syntax given in template.tex or provide your algorithm as a figure.
 #'   \item Please do not define new commands.
-#'   \item The most commonly used packages (\code{\\usepackage{}}) are integrated in the copernicus.cls. Some other packages often used by the community are defined in template.tex. Please do not insert additional ones in your *.tex file.
+#'   \item The most commonly used packages (\code{\\usepackage{}}) are integrated in the \code{copernicus.cls}. Some other packages often used by the community are defined in template.tex. Please do not insert additional ones in your *.tex file.
 #'   \item Spaces in labels (\code{\\label{}}) are not allowed; please make sure that no label name is assigned more than once.
 #'   \item Please do not use \code{\\paragraph{}}; only \code{\\subsubsection{}} is allowed.
 #'   \item It is not possible to add tables in colour.
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -56,10 +56,12 @@ copernicus_article <- function(
     "-auto_identifiers"    # disables \hypertarget commands
   )
 ) {
-  if("extra_dependencies" %in% names(list(...)))
-    warning("Copernicus does not support additional LaTeX packages and options!
+  if ("extra_dependencies" %in% names(list(...)))
+    warning(
+      "Copernicus does not support additional LaTeX packages and options!
           >> Please remove 'extra_dependencies' from your YAML header!",
-         call. = FALSE)
+      call. = FALSE
+    )
 
   pdf_document_format(
     "copernicus", citation_package = citation_package,
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -26,7 +26,9 @@
 #'   \item Please use only commands in which words, numbers, etc. are within braces (e.g. \code{\\textrm{TEXT}} instead of \code{{\\rm TEXT}}).
 #'   \item For algorithms, please use the syntax given in template.tex or provide your algorithm as a figure.
 #'   \item Please do not define new commands.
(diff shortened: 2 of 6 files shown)


inst/rmarkdown/templates/copernicus/resources/template.tex
```diff
@@ -111,11 +111,6 @@
 %$header-includes$
 %$endfor$
 
-$if(algorithms)$
-\usepackage{algorithmic}
-\usepackage{algorithm}
-$endif$
```
cderv, line 117: Removing this and the associated yaml option will be a breaking change for all current users who sets it. 

Is it not part of the Copernicus template anymore ? 

I believe we can leave that in the template and suggest to set it to false by default in the YAML header ? 

Which would that not work ?

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
```diff
@@ -46,8 +46,6 @@ running:
 # This section is mandatory even if you declare that no competing interests are present.
 competinginterests: |
   The authors declare no competing interests.
-# OPTIONAL:
-algorithms: true
```
cderv, line 50: This could be set to FALSE by default here so that it is not included unless opt-in ?

Is there a problem by doing that ?

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
```diff
@@ -290,10 +290,13 @@ x & y & z\\
...
 $$
 
-## ALGORITHM
+## ALGORITHM/PROGRAMMING CODE
 
-If you want to use algorithms, you can either enable the required packages in the header (the default, see `algorithms: true`), or make sure yourself that the \LaTeX packages `algorithms` and `algorithmicx` are installed so that `algorithm.sty` respectively `algorithmic.sty` can be loaded by the Copernicus template.
-Copernicus staff will remove all undesirable packages from your LaTeX source code, so please stick to using the header option, which only adds the two acceptable packages.
+If you want to use algorithms or programming code, you can either use the default 'rmarkdown' functionality, which includes the
+code in \LaTeX{} using the `verbatim` environment or using the \LaTeX{} packages `algorithms` and `algorithmicx`. Both need 
+to be available through your preferred \LaTeX{} distribution.
```
cderv, line 297: So are the two LaTeX package no more accepted by the Copernicus official template and that is why there are removed ?  

Otherwise, the current switch with YAML field is enough, isn't it ? 

I am not quite sure to see the issue here.

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
```diff
@@ -290,10 +290,13 @@ x & y & z\\
...
-Copernicus staff will remove all undesirable packages from your LaTeX source code, so please stick to using the header option, which only adds the two acceptable packages.
+If you want to use algorithms or programming code, you can either use the default 'rmarkdown' functionality, which includes the
+code in \LaTeX{} using the `verbatim` environment or using the \LaTeX{} packages `algorithms` and `algorithmicx`. Both need 
+to be available through your preferred \LaTeX{} distribution. 
+
+**Please note:** Copernicus does not support additional \LaTeX{} packages or new \LaTeX{} commands. This extends to syntax highlighting of source code. Therefore this template sets the parameter `highlights` in the YAML header to `NULL`. However, it might be desirable to have syntax highlight available in preprints or for others reasons. Please see `?rmarkdown::pdf_document` for available options.
```
cderv, line 299: Yes setting highlight to NULL will prevent Pandoc to insert its macros and highlighting code. 👍

R/copernicus_article.R
```diff
@@ -12,21 +12,21 @@
...
 #'   \item Please use only commands in which words, numbers, etc. are within braces (e.g. \code{\\textrm{TEXT}} instead of \code{{\\rm TEXT}}).
 #'   \item For algorithms, please use the syntax given in template.tex or provide your algorithm as a figure.
 #'   \item Please do not define new commands.
-#'   \item The most commonly used packages (\code{\\usepackage{}}) are integrated in the copernicus.cls. Some other packages often used by the community are defined in template.tex. Please do not insert additional ones in your *.tex file.
+#'   \item The most commonly used packages (\code{\\usepackage{}}) are integrated in the \code{copernicus.cls}. Some other packages often used by the community are defined in template.tex. Please do not insert additional ones in your *.tex file.
 #'   \item Spaces in labels (\code{\\label{}}) are not allowed; please make sure that no label name is assigned more than once.
 #'   \item Please do not use \code{\\paragraph{}}; only \code{\\subsubsection{}} is allowed.
 #'   \item It is not possible to add tables in colour.
```
cderv, line 32: Maybe a rule can be added here about the highlighting stuff: 

* Do not forget to use `highlight = NULL` for editor review. It has been set to NULL to not forget it. It can be set to other value during preprint step for example for conveniance.

It is what I understood.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 rticles 0.20
 ---------------------------------------------------------------------
+- Fixes a problem in the Copernicus Publications template that causes a manuscript to bounce back to the author during the typesetting step. This was triggered by the inclusion (unknown to the author) of non-supported LaTeX packages and commands if programming code was part of the manuscript using the standard 'rmarkdown' code support. Additionally, the template gained support for the `highlight` parameter of `rmarkdown::pdf_document` to enable or disable syntax 
+highlight (disabled by default, as required by Copernicus)
```
cderv, line 4: About the new argument `highlight`, I think it should be highlighted in the NEWS that this is a breaking change because there won't be any highlighting by default now. One as to set `highlight = "default"` to get the previous highlighting. 

Just so other users are not surprised by the change even if it is justified 😅

cderv (changes requested):
Thanks a lot for this feedback ! 

That is really great that as users you can keep these templates up to date with the editor requirement. 

I have a few comment and questions about the change. 

I understood the highlight one but not sure to see why setting `algorithms: false` in the YAML is not enough when you want to deactive algorithm environment. 
Removing this will be a breaking change also, even if it is justified by your discussion with editor.

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: Yeah, this indeed a little bit confusing. In fact, the two LaTeX packages are still used by Copernicus, and *already* imported through the `copernicus.cls`. However, if we include them in the way we are currently doing it the `*.tex` will come back from the journal's office.

inst/rmarkdown/templates/copernicus/resources/template.tex
RLumSK: Not necessarily, because the `copernicus.cls` imports them either way. It was basically doubled in the `*.tex`, therefore Copernicus sent the manuscript back and ask to remove it.

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: Same as above. But I see your point that there is a very slim chance that something gets somehow broken, but it does not make much sense to produce files that would not be accepted by the publisher.

NEWS.md
RLumSK: Good point. I added this to the news.

R/copernicus_article.R
RLumSK: This is an important point, thank you. I added more information to the documentation as requested.

inst/rmarkdown/templates/copernicus/resources/template.tex
cderv: Oh you mean it is included in the cls file by copernicus already ?

I did not check but yes it is ! 
https://github.com/rstudio/rticles/blob/4aa0a44e4f41b02cfb1d5c1cdc80043b5f42e502/inst/rmarkdown/templates/copernicus/skeleton/copernicus.cls#L1681-L1711

so that is enough if you have the package installed ! 

I don't think **tinytex** will automatically install them though if missing - which is different than when you use `\usepackage`

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
cderv: I see now. 

I just think the way it is done by the cls file will prevent **tinytex** to install the package automatically if missing. 

So algorithm won't work in the skeleton unless installed manually by the user. This goes a bit against the use of **rmarkdown** + **tinytex** to compile a document without worrying about CTAN packages. 

by adding the `\usepackage` in template this was triggering the installation if missing. 

Should be install those packages in `copernicus_article` because they are needed anyway by the cls file to use all feature ? 

Or maybe in the setup chunk of the skeleton ? 

I think we have already done that with a template for the same reason: a cls file while loading a package if already installed. But this means that if not, **tinytex** won't see any error to trigger an installation. 

https://github.com/rstudio/rticles/blob/10d269397fe3af52ffeebfab533a3eb970567f33/inst/rmarkdown/templates/jasa/skeleton/skeleton.Rmd#L59-L61

R/copernicus_article.R
```diff
@@ -12,21 +12,23 @@
...
+#' Copernicus converts all typeset TeX files into XML, the expressions and markups have to be highly standardized.
 #' Therefore, please keep the following in mind:
 #'
 #' \itemize{
 #'   \item Please provide only one figure file for figures with several panels, and please do not use \code{\\subfloat} or similar commands.
 #'   \item Please use only commands in which words, numbers, etc. are within braces (e.g. \code{\\textrm{TEXT}} instead of \code{{\\rm TEXT}}).
 #'   \item For algorithms, please use the syntax given in template.tex or provide your algorithm as a figure.
 #'   \item Please do not define new commands.
-#'   \item The most commonly used packages (\code{\\usepackage{}}) are integrated in the copernicus.cls. Some other packages often used by the community are defined in template.tex. Please do not insert additional ones in your *.tex file.
+#'   \item Supported packages (\code{\\usepackage{}}) are already integrated in the \code{copernicus.cls}.  Please do not insert additional ones in your \code{*.tex} file.
```
cderv, line 29: OK so now I see: this is the main change ! 

Copernicus does not allow more package to be used - only those already define in the cls file, is it ? 

How do you know which package can be used then ? because I supposed you need them to be installed and unfortunately copernicus.cls is not available in a CTAN package that would also install dependency. 

How can we help user know about that from the Rmd skeleton or the doc ? 

by the way, if this is the rule, you could warn more about that in `copernicus_article()` bc it means `extra_dependencies:` is not allowed as it will add `\usepackage` in preamble. 

Could be a way to avoid later surprises ? 

Also maybe we should add a comment in the rticles's template so that further contributor know that there should be no `\usepackage` command added in this template

What do you think ? 

cc @nuest

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: Update: this was probably not an issue in the past. However, according to what the publisher wrote to us, the way they are currently processing manuscripts is that the LaTeX code is translated into XML and then everything along the line is processed from there. It appears that their parser is not really forgiving. So probably it produced an error and before fixing that manually for every manuscript, they just send it back to the authors.

cderv: yes I now see that it will still be used but the package needs to be installed already in the user system to avoid further error if `\begin{algorithm}` is used.

inst/rmarkdown/templates/copernicus/resources/template.tex
RLumSK: Yes, I do agree that **tinytex** would be of no help here and cannot install missing packages automatically. On the other hand, the CLS uses also other (more commonly used; admittedly) packages. But it appears that the build process would just continue if those packages are missing in the available LaTeX installation.

R/copernicus_article.R
RLumSK: @cderv, Thank you for your detailed feedback. I always realise again that you have a broader picture in mind than me. 

> Copernicus does not allow more package to be used - only those already define in the cls file, is it ?

Yes, according to @nuest this was probably allowed in the past. I cannot tell. 

> How do you know which package can be used then ? because I supposed you need them to be installed and unfortunately copernicus.cls is not available in a CTAN package that would also install dependency.

Oh, Daniel found already a good solution, I think, in the `template.tex` he had added all supported packages:

```%% \usepackage commands included in the copernicus.cls:
%\usepackage[german, english]{babel}
%\usepackage{tabularx}
%\usepackage{cancel}
%\usepackage{multirow}
%\usepackage{supertabular}
%\usepackage{algorithmic}
%\usepackage{algorithm}
%\usepackage{amsthm}
%\usepackage{float}
%\usepackage{subfig}
%\usepackage{rotating}
```

> How can we help user know about that from the Rmd skeleton or the doc ?

Perhaps the cheap solution is adding this information from the `template.tex` to the `*.Rmd` file?

> by the way, if this is the rule, you could warn more about that in copernicus_article() bc it means extra_dependencies: is not allowed as it will add \usepackage in preamble.

Thanks, I added this to the documentation and I added a `stop()` to the `copernicus_article()` if the user tries something that would not be allowed.. I guess we would not a test too, but here I need your help. I have no idea where I should add this unit test.

>Could be a way to avoid later surprises ?

As written above, we could just add the list of supported packages to the `*.Rmd` file. Perhaps that is enough?

> Also maybe we should add a comment in the rticles's template so that further contributor know that there should be no \usepackage command added in this template

A valuable point. I added such a piece of information to the `template.tex`.

nuest: Yes, we should add a comment in the template main body to make sure users see that.

Good point about `extra_dependencies` - do you suggest to catch that in the function and put out a `warning()` ?

inst/rmarkdown/templates/copernicus/resources/template.tex
cderv: > But it appears that the build process would just continue if those packages are missing in the available LaTeX installation.

And it means it will fail if you use one of the command from those packages without having installing it ? 

I guess this is usual LaTeX user experience. We could prevent that if we have a way to know which are the packages expected, otherwise we can't really I think

nuest: @cderv Just to make it clear to me: **tinytex** will not recognise the `\RequirePackage{..}` in the CLS file?

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: Do you suggest I should implement this in the skeleton for all packages? Might be a solution indeed.

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: Agreed!

nuest: Maybe the most straightforward solution would be to mention that `algorithm` and/or `algorithmx` have to be installed when the environment is used, _in the part of the template where algorithms are shown?_

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
nuest: Ah, strike my comment above :-) If it's alright to do this in the skeleton, I think that is a good idea - it won't be needed for users without algorithms, but it won't really hurt them either.

R/copernicus_article.R
RLumSK: Actually, I already added a hard `stop()` in the code with feasible advice to the user. But perhaps the `stop()` is too much?

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
cderv: I think this could be put in the skeleton in the setup chunk, at least for this two packages that we know needs to be install for the skeleton to work. 

We could indeed add this for all packages in your list
```r
if (tinytex::is_tinytex()) {
	local({
		  required <- c(...)
		  installed <- tinytex::check_installed(required)
		  tinytex::tlmgr_install(required[!installed])
	})
}
```

One could remove it if it is not desired.

R/copernicus_article.R
cderv: yes maybe stop is too much... 🤔 
Maybe just a warning as this is really advice to follow if you want to submit your paper to copernicus ? 

We could also add explicitly the argument to the function, to documented it in the help page;, catch a different value than NULL, and stop or warn that it will be set to NULL. Usually, if you have extra dependencies, it will be required to compile so a stop it could be. 

I don't know who would use this template but not published with the editor...

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
cderv: Yes I think it should at least be mention in the part where it is used in the skeleton, maybe by showing a chunk with `include = FALSE` to install those packages so the at least the skeleton works. don't you think ?

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: Stupid question: Is there any argument against implementing this directly in the `copernicus_article.R`? If there is, I would add it to the `*.Rmd` file instead as suggested.

R/copernicus_article.R
RLumSK: Ok, it is now a `warning()`. I did not add another argument, because it would add something, which is basically not wanted anyway by the publisher (means, better keeping documentation concise).

inst/rmarkdown/templates/copernicus/skeleton/skeleton.Rmd
RLumSK: As written below, I guess we could circumvent this if checking and loading additional packages is allowed in the `copernicus_article.R`. @cderv the decision is all yours. I am ok with both solutions, but consider the integration in `copernicus_article.R` as more elegant.

cderv (approved):
I did take into account the comments and tweaked a little. 

After some tests and more thinking, now I think that is is not a good idea to try install from the output format function directly. 

First, it is hard to know which LaTeX distribution is in used. We can only provide the command for TinyTeX or for TeX Live, but users needs write access for this. So it can be the source of more unknown error from user to try doing it automatically. I think it is best that a user do the installation by itself. 
Also, it is kind of hard to keep track of the required CTAN package to install to go with the `\usepackage` clauses in the `template.tex`. Some are missing and some  are not exactly the name of their package. I don't want to try maintaining a list we don't have a true source of. 

This is why I added documentation in the skeleton about all the feedback you have,  including the need to install the two packages for algorithms. I added the command  to run with **tinytex** for user and included a chunk to install them that is evaluated only when TinyTeX distribution is used. 

I think this will help as much as we could. Hope it is fine with you. 

We can still modify later if I missed something.

RLumSK:
Great and thanks for your support all fine with me!

cderv:
Thanks for your contribution. Those templates works thanks to you all !

PR #393, opened 2021-04-27
Opt-out bookdown environment definition for lipic article

### LINKED ISSUE #392: resolve conflict between lipic format and bookdown regarding environment
This follows https://github.com/rstudio/rticles/pull/288#issuecomment-721232934 and https://github.com/rstudio/bookdown/issues/1001

and the new option in dev **bookdown** : `bookdown.theroem.preamble`

We need to set that to FALSE for some formats where there is a conflict and document

cc @nuest

cderv:
This fixes #392 (cc @nuest)

This can now be used in YAML 
```
output:
  bookdown::pdf_book:
    base_format: rticles::lipics_article
```
and it will no more error with 
```
! LaTeX Error: Command \theorem already defined.
               Or name \end... illegal, see p.192 of the manual.
``` 

This change set an option for opting out the addition by **bookdown** of some definition. This requires **bookdown** 0.23 (currently dev version) to work

```r
remotes::install_github("rstudio/bookdown")
```

@yihui 
 * is this how you would do it ? I used the `on_exit` argument to be sure to reset the options after `render()`
 * I did not set a requirement for **bookdown**  up in DESCRIPTION as it is in suggest, should I ? 
 * This requires **bookdown** 0.23, should we merge in **rticles** only when it is release ? I added a mention in NEWS about the requirement for this.

### DIFF
--- a/R/article.R
+++ b/R/article.R
@@ -252,23 +252,11 @@ lipics_article <- function(
 ) {
   # quick dev shortcut for Ubuntu: click "Install and restart" then run:
   # unlink("MyArticle/", recursive = TRUE); rmarkdown::draft("MyArticle.Rmd", template = "lipics", package = "rticles", edit = FALSE); rmarkdown::render("MyArticle/MyArticle.Rmd"); system(paste0("xdg-open ", here::here("MyArticle", "MyArticle.pdf")))
-  base <- pdf_document_format(
+  pdf_document_format(
     "lipics", latex_engine = latex_engine,
     citation_package = citation_package, keep_tex = keep_tex,
     md_extensions = md_extensions, ...
   )
-
-  # we deactivate bookdown addition of environment definitions in preamble
-  # to avoid conflict as lipics already defines some
-  opts <- options(bookdown.theorem.preamble = FALSE)
-  on_exit <- function() options(opts)
-
-  base_on_exit <- base$on_exit
-  base$on_exit <- function() {
-    if (is.function(base_on_exit)) base_on_exit()
-    on_exit()
-  }
-  base
 }
 
 #' @section \code{mdpi_article}: Format for creating submissions to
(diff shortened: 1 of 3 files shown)


R/article.R
```diff
@@ -252,11 +252,23 @@ lipics_article <- function(
...
+  # to avoid conflict as lipics already defines some
+  opts <- options(bookdown.theorem.preamble = FALSE)
+  on_exit <- function() options(opts)
+
+  base_on_exit <- base$on_exit
+  base$on_exit <- function() {
+    if (is.function(base_on_exit)) base_on_exit()
+    on_exit()
+  }
+  base
```
yihui, line 271: That sounds too complicated. I'd just add a single line of `options(bookdown.theorem.preamble = FALSE)` to `skeleton.Rmd` (e.g., in the `setup` chunk).

No strong opinion, though. You can decide by yourself.

yihui (approved):
> * I did not set a requirement for **bookdown**  up in DESCRIPTION as it is in suggest, should I ?

Yes, even though version requirements in Suggests are actually ignored when loading packages. Having it in DESCRIPTION is at least a reminder to users.

> * This requires **bookdown** 0.23, should we merge in **rticles** only when it is release ?

No need to wait. We can merge now and let @nuest test it with the dev version of **bookdown**.

Thanks!

R/article.R
cderv: Oh ok. I recognize this is fairly complicated - It took me some time to write this correctly :)

but I did not want to expose this "kind of internal" option to users. I also choose to put that inside the output function so that previous articles with old Rmd template would benefit from it when using new bookdown article. 

Setting on exit hook to `render()` is coming more and more in our usage. Currently doing this through the format is the only way I know to do it properly. Maybe we are missing helper function to that properly. `rmarkdown:::merge_on_exit()` is the one that would simplify currently 

```r
opts <- options(bookdown.theorem.preamble = FALSE)
base$on_exit <- rmarkdown:::merge_on_exit(base$on_exit, function() options(opts))
```

There may be other way. 

Anyway, I am fine also by exposing this in a setup chunk for the articles skeleton.

cderv: I decided to follow you advice on this one and simplify this. Thanks !

nuest:
:tada:

PR #402, opened 2021-05-17
Expose biblio-style in elsarticle

gregmacfarlane:
This is a minor commit that exposes the bookdown `biblio-style` argument to the `elsarticle` LaTeX template when using natbib.

### DIFF
--- a/inst/rmarkdown/templates/elsevier/resources/template.tex
+++ b/inst/rmarkdown/templates/elsevier/resources/template.tex
@@ -57,11 +57,7 @@ $if(geometry)$
 $endif$
 $if(natbib)$
 \usepackage{natbib}
-$if(biblio-style)$
-\bibliographystyle{$biblio-style$}
-$else$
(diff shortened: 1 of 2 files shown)


inst/rmarkdown/templates/elsevier/resources/template.tex
```diff
@@ -57,7 +57,11 @@
 $endif$
 $if(natbib)$
 \usepackage{natbib}
+$if(biblio-style)$
+\bibliographystyle{$biblio-style$}
+$else$
 \bibliographystyle{plainnat}
+$endif$
```
cderv, line 64: I'll just do it in one liner like in Pandoc template
```suggestion
\bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
```

cderv (approved):
Thanks for this ! 

I'll update NEWS accordingly

Issue #403, opened 2021-05-21
Fix issue with mnras format 


cderv:
Seems like there is an issue with the mnras rticles format 

https://github.com/rstudio/rticles/runs/2637643070?check_suite_focus=true#step:15:229

cderv:
Issue is 

````
! Incomplete ifx; all text was ignored after line 3.
<inserted text> 
                fi 
l.112 ...\familydefault\seriesdefault\shapedefault
````

**Comment by cderv** [comment]
I am not sure this comes from our template. mnras files are from CTAN : https://www.ctan.org/tex-archive/macros/latex/contrib/mnras

@oleskiewicz pinging you in case you know better as you contributed the template in the first place I believe. 

I am not sure to know where this error comes from in the tex file.

**Comment by oleskiewicz** [comment]
When I added the template, I aimed to faithfully translate the example from:
http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/mnras/, I'm not aware of any 
changes to the template since.

I took a look, but I can't figure out what the error is, or why it started failing now.

I'll try reproducing it locally in a fresh environment and sending a PR if I manage to 
find a fix.

**Comment by oleskiewicz** [comment]
I managed to build a new article with

```r
rmarkdown::draft("article.Rmd", template = "mnras", package = "rticles")
rmarkdown::render("./article/article.Rmd")
```

but I had to add http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/mnras/mnras.cls file to `./article/` directory, and manually delete `vadjust` environment (this I believe was solved by #380.  So I think we can just add `mnras.cls` file to the skeleton directory and that should fix it!

cderv:
Thanks @oleskiewicz ! 

`mnras.cls` is not included into **rticles** because it should be available on CTAN (which is where you pointed to) and installed by **tinytex** when the Rmd is converted.
````
> tinytex::parse_packages(files = "mnras.cls")
tlmgr search --file --global "/mnras.cls"
[1] "mnras"
````

I'll look into this.

cderv:
@oleskiewicz I tested a wild thoughts: would the issue have fixed itself... 
So I triggered a recheck of the actions to see if there would still be an error and it seems not. See #408 -> every checks are green ✔️ ! 

It means that it could have been an issue with the mnras package not being properly fetch the first time. 

This now works for me on my side
````r
withr::with_tempdir({
[...]

PR #407, opened 2021-05-29
add mnras cls file (#403)

oleskiewicz:
this PR hopefully fixes #403

PR #408, opened 2021-05-31
trigger GitHub actions

cderv:
Just to check if #403 is fixed by itself following last comment in issue thread


cderv:
Seems like check are passing now and nothing has to be done to close #403 - #407 may not be needed.

cderv:
Nothing to do then.

PR #418, opened 2021-07-05
Update to oup_article for new CTAN template

dmkaplan2000:
I have updated the OUP format to use the latest template available on CTAN. I do, however, have a number of comments regarding the update:

1) I didn't include the .csl file as they are on CTAN. However, as what is on CTAN doesn't include a .sty file, I am not 100% sure that the style is easily installable with tlmgr. I don't use tinytex, so it is hard for me to check if that is the case or not.
2) This new template is largely backward compatible with the old template, but there was no simple way to maintain full backward compatibility. In particular, multi-paragraph abstracts in the old format won't work in the new.
3) I discovered that not all OUP Journals currently accept this new format. In particular, one that I submit to, ICES Journal of Marine Science, does not use this format. As such, I will submit a separate pull request creating a format for this specific journal that is still based on the old OUP style. 

-----------------------------------------------------------

- [X ] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [X ] Add a skeleton article `inst/rmarkdown/templates/journalname/skeleton/skeleton.Rmd`.

- [X ] Add a description of the template `inst/rmarkdown/templates/journalname/template.yaml`.

- [ X] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. If you are using TinyTeX or TeX Live, you can verify if a package is available on CTAN via `tinytex::parse_packages(files = "FILENAME"")` (e.g., when `FILENAME` is `plain.bst`, it should return `"bibtex"`, which means this file is from a standard CTAN package). Please keep the number of new files absolutely minimal (e.g., do not include PDF output files), and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include too many items in it without using all of them).

- [ X] Update Rd and namespace (could be done by `devtools::document()`).

- [ X] Update NEWS.

- [ X] Update README with a link to the newly supported journal. Please add your Github username and the full name of the journal (follow other examples in the list). 

- [ X] Add a test to `tests/testit/test-formats.R` by adding a line `test_format("journalname")`. We try to keep them in alphabetical order.

- [X ] Add your name to the list of authors `Authors@R` in DESCRIPTION. You don't need to bump the package version in DESCRIPTION.

Thank you!


dmkaplan2000:
> x David M. Kaplan

David M. Kaplan" is from some old commits I did predating this branch where I used my name as the git commit email address. All commits since branching that are relevant to this pull request have been done with my github email address, for which the CLA has been signed. I haven't found a convenient way to fix the commit address on the old commits predating this branch.

dmkaplan2000:
I looked into the R-CMD-check errors and they appear to be due to older versions of pandoc not understanding certain elements of my template.tex (specifically, pandoc 2.7.3 does not understand $elseif()$ and pandoc 2.9.2.1 does not seem to know about the pipe "first"). The development version of pandoc also produces an error, but that is not related to the OUP template. The strange thing is that on my system running Ubuntu 20.04 with pandoc 2.5, everything works fine. 

In any case, should I be trying to fix these issues? `$elseif$` could be fixed if need be (it would be messy, but presumably it could be done), but replacing the pipe `first` would be a bit more complicated I think.

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -27,15 +50,14 @@
...
-
-% Packages that are supposedly required by OUP sty file
-\usepackage{amssymb, amsmath, geometry, amsfonts, verbatim, endnotes, setspace}
+\usepackage{float}
+\usepackage{colortbl}
+\usepackage{pdflscape}
+\usepackage{tabu}
+\usepackage{threeparttable}
 
 % For code highlighting I think
```
cderv, line 62: I think we need this instead: 
````
$if(highlighting-macros)$
$highlighting-macros$
$endif$
```` 

See https://github.com/rstudio/rticles/pull/419#discussion_r678165258

File: inst/rmarkdown/templates/oup/resources/template.tex [hunk 2]
```diff
@@ -96,48 +122,76 @@
...
+    {}
+    {\refstepcounter{correspcnt}\newcounter{CORRESPLBL@#1}\setcounter{CORRESPLBL@#1}{\value{correspcnt}}}%
+    \fnsymbol{CORRESPLBL@#1}%
+}
+
+\newcommand*\correspbycode[1]{%
+  \ifcounter{CORRESP@#1}
+    {}
+    {\newcounter{CORRESP@#1}%
+     \corresp[\corresplblbycode{#1}]{\MyRef{CORRESPTXT}{#1}}}%
```
cderv, line 150: About all the above here: Is this custom for backward compatibility ? Is it from template ? 

Trying to understand to facilitate maintainance of this in the future

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -96,48 +122,76 @@
...
+  $if(address.state)$\state{$address.state$}, $endif$%
+  $if(address.country)$\country{$address.country$}$endif$%
+  $if(address.street)$}%
+   $elseif(address.city)$}%
+   $elseif(address.postcode)$}%
+   $elseif(address.state)$}%
+   $elseif(address.country)$}%
+  $endif$%
+ $endif$%
+ }
```
cderv, line 186: This is a breaking change for previous user if they rerun their article using `oup_article()` ?

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -169,72 +223,129 @@
...
 
 \begin{document}
 
-\title{$title$}
+\journaltitle{$if(journal)$$journal$$else$Journal Title Here$endif$}
+\DOI{DOI HERE}
+\copyrightyear{YYYY}
+\pubyear{YYYY}
+\access{Advance Access Publication Date: Day Month Year}
+\appnotes{Paper}
```
cderv, line 239: Does this need to be customized by user ? 

If so, it needs to be available through variables to set in the YAML header.

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -169,72 +223,129 @@
...
+$elseif(authors)$
+\authormark{$for(authors/first)$$it.name$$endfor$$if(authors/rest/rest)$ et al.$else$$for(authors/rest)$ \and $authors.name$$endfor$$endif$}
+$endif$
+
+\received{Date}{0}{Year}
+\revised{Date}{0}{Year}
+\accepted{Date}{0}{Year}
 
-\date{$if(date)$$date$$else$\today$endif$}
+%\editor{Associate Editor: Name}
```
cderv, line 300: Should it be available to modified by user also ?

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -169,72 +223,129 @@
...
 $body$
 
-$if(notes)$
-\begin{notes}
-$notes$
-\end{notes}
+%%%%%%%%%%%%%%
+
+%\begin{appendices}
+%\end{appendices}
```
cderv, line 331: Should a solution for this needs to be found ? 

It is possible that this could be added at the end of the rmarkdown file this way 

````markdown

::: {.appendices latex=true}

# Section title of first appendix

blabla

## Subsection title of first appendix

and so ....
:::
````

File: inst/rmarkdown/templates/oup/resources/template.tex [hunk 7]
```diff
@@ -169,72 +223,129 @@
...
+
+$if(competing_interests)$
+\section{Competing interests}
+
+$competing_interests$
 $endif$
```
cderv, line 337: This could also be but in end of body as usual markdown and the other below also. 
Anything before bibliography insertation can be done this way. 

````
# Competing interests

There is NO Competing Interest

# Author contributions statement

Must include all authors, identified by initials, for example:
S.R. and D.A. conceived the experiment(s),  S.R. conducted the experiment(s), S.R. and D.A. analysed the results.  S.R. and D.A. wrote and reviewed the manuscript.

# Acknowledgments

The authors thank the anonymous reviewers for their valuable suggestions. This work is supported in part by funds from the National Science Foundation (NSF: \# 1636933 and \# 1920920).
````

But using YAML header is also fine.

File: inst/rmarkdown/templates/oup/resources/template.tex [hunk 8]
```diff
@@ -248,13 +359,22 @@
...
 $endif$
 
+\bibliographystyle{$if(namedate)$abbrvnat$else$plain$endif$}
```
cderv, line 362: In usual pandoc template we have 
```suggestion
\bibliographystyle{$if(biblio-style)$$biblio-style$$else$plain$endif$}
```

And it would be in YAML 

````yaml
bibliography: mybib.bib
# biblio-style: abbrvnat # uncomment line to use AUTHOR YEAR FORMAT, or change value to change style
````

I would prefer to keep this way, following generic Pandoc doc rather than having a specific variable `namedate` for this format

File: inst/rmarkdown/templates/oup/resources/template.tex [hunk 9]
```diff
@@ -96,48 +122,76 @@
...
+\MyNewLabel{CORRESPTXT}{$footnote.code$}{$footnote.text$}
 $endif$
 $endfor$
-$endif$
-
-% Part for setting citation format package: natbib
-$if(natbib)$
-\usepackage{natbib}
-\bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
-$endif$
```
cderv, line 130: Why remove this ? 

We still need `\usepackage{natbib}`, don't we ?

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
```diff
@@ -23,69 +23,75 @@ author:
...
+output:
+  rticles::oup_article:
+    keep_tex: TRUE
+    journal: "Journal Name"
+    #document_style: "contemporary" # Can be contemporary, modern, traditional; default "contemporary"
+    #papersize: "large" # Can be large, medium, small; default large
+    #citation_package: "default" # Uncomment if using a CSL biblio style; default "natbib"
+    namedate: TRUE # Set FALSE to use numeric refs; Default FALSE
+    #number_sections: FALSE # Uncomment to not number sections; default TRUE
+    #onecolumn: TRUE # Uncomment for one column format; default FALSE
```
cderv, line 74: All this execept `number_sections` could be set as first level key in YAML header. 
What advantage to put them in the function format ?

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
```diff
@@ -23,69 +23,75 @@ author:
...
 #urlcolor: orange
 #linkcolor: green
 #citecolor: red
 #header-includes:
 #  - \usepackage[nomarkers,tablesfirst]{endfloat} # For figures and tables at end
-#  - \usepackage{lineno} # For line numbering
-#  - \linenumbers # For line numbering
+#  - \usepackage{lineno}\linenumbers # For line numbering
```
cderv, line 80: I think you put that already int the template using  using a variable. 
Maybe this option should be shown ? instead of header includes.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
```diff
@@ -100,51 +106,118 @@ A bullet list:
...
 ## Generate a table using `kable`
 
-```{r anothertable, echo=TRUE}
+```{r kabletable, echo=TRUE}
 df = data.frame(ID=1:3,code=letters[1:3])
 
 # kable can alse be used for creating tables
-knitr::kable(df,caption="This is the table caption",format="latex",
+knitr::kable(df,caption="This is a kable table.",
+             #format="latex",
```
cderv, line 191: If you add the option, you should remove the comment here. 
Otherwise, don't add the option and show how to use a format argument.

`kable()` can output also Markdown table, that will be converted to LaTeX by Pandoc and this should work too.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
```diff
@@ -23,69 +23,75 @@ author:
...
   This is an acknowledgement.
 
   It consists of two paragraphs.
-keywords:
-  - key
-  - dictionary
-  - word
-#fontsize: 12pt
-#spacing: halfline # could also be oneline
 #classoptions:
```
cderv, line 61: Should the be example of classoptions to set ? 

The bundled template with OUP package seems to show several examples 

````latex
%%%CONTEMPORARY%%%
\documentclass[unnumsec,webpdf,contemporary,large]{oup-authoring-template}%
%\documentclass[unnumsec,webpdf,contemporary,large,namedate]{oup-authoring-template}% uncomment this line for author year citations and comment the above
%\documentclass[unnumsec,webpdf,contemporary,medium]{oup-authoring-template}
%\documentclass[unnumsec,webpdf,contemporary,small]{oup-authoring-template}

%%%MODERN%%%
%\documentclass[unnumsec,webpdf,modern,large]{oup-authoring-template}
%\documentclass[unnumsec,webpdf,modern,large,namedate]{oup-authoring-template}% uncomment this line for author year citations and comment the above
%\documentclass[unnumsec,webpdf,modern,medium]{oup-authoring-template}
%\documentclass[unnumsec,webpdf,modern,small]{oup-authoring-template}

%%%TRADITIONAL%%%
%\documentclass[unnumsec,webpdf,traditional,large]{oup-authoring-template}
%\documentclass[unnumsec,webpdf,traditional,large,namedate]{oup-authoring-template}% uncomment this line for author year citations and comment the above
%\documentclass[unnumsec,namedate,webpdf,traditional,medium]{oup-authoring-template}
%\documentclass[namedate,webpdf,traditional,small]{oup-authoring-template}
````

**Review by cderv — CHANGES_REQUESTED** [review]
Thanks for this important update. 

My main concerns is with breaking change. 

The update seems important and maybe this could cause issue to user still using the old cls file. 
All the more if some journal does not accept yet the new style. 

So I am hesitant on what to do exactly: 

* Create a new function 
* Have a switch is this function to use the old or new template ? We would have both and copy / use the correct one depending on what is asked. We could still have the old as default maybe but this could allow us to deprecate the function and inform of changes before changing the default. 

I am not quite sure of the difference however and it seems this could be minimum. 

> I didn't include the .csl file as they are on CTAN. However, as what is on CTAN doesn't include a .sty file, I am not 100% sure that the style is easily installable with tlmgr. I don't use tinytex, so it is hard for me to check if that is the case or not.

Do you mean `cls` file  ? The new file will be found so we are ok
````
> tinytex::parse_packages(files = "oup-authoring-template.cls")
tlmgr search --file --global "/oup-authoring-template.cls"
[1] "oup-authoring-template"
```` 
If you meant CSL, where is it available , 

> This new template is largely backward compatible with the old template, but there was no simple way to maintain full backward compatibility. In particular, multi-paragraph abstracts in the old format won't work in the new.

I understand that my concerns are to be minimized because you are saying this is backward compatible, is that right ? Do you know the list of changes that we could document / inform ? 

> I discovered that not all OUP Journals currently accept this new format. In particular, one that I submit to, ICES Journal of Marine Science, does not use this format. As such, I will submit a separate pull request creating a format for this specific journal that is still based on the old OUP style.

This could mean also that ICES JMS may not be the only one. is `icesjms_article` required if you were able to still use somehow the current `oup_rticles()` function ? 


Thanks again for you valuable contribution !

**Comment by cderv** [comment]
> I looked into the R-CMD-check errors and they appear to be due to older versions of pandoc not understanding certain elements of my template.tex (specifically, pandoc 2.7.3 does not understand $elseif()$ and pandoc 2.9.2.1 does not seem to know about the pipe "first"). The development version of pandoc also produces an error, but that is not related to the OUP template. The strange thing is that on my system running Ubuntu 20.04 with pandoc 2.5, everything works fine.

For the CI error, it would require that master should be merged. There are some fixes already for some error in other template. As of now, all templates are tested in a PR and not just the one you change. 

About the pandoc requirement, yes template syntax is dependant of Pandoc version. 

* Pipes where introduces in Pandoc 2.10 https://pandoc.org/releases.html#pandoc-2.10-2020-06-29
* `elseif` in Pandoc 2.8: https://pandoc.org/releases.html#pandoc-2.8-2019-11-22

It should not work in Pandoc 2.5 for you. 

We could make a minimal requirement on Pandoc for a format. I am open to that. However, it would justify a new format or an opt-in way to not impact user of current `oup_article`

**Comment by dmkaplan2000** [comment]
> Thanks for this important update.
> 
> My main concerns is with breaking change.
> 
> The update seems important and maybe this could cause issue to user still using the old cls file.
> All the more if some journal does not accept yet the new style.

Yes, I agree with this, but haven't figured out the best solution. I talk about some of the issues involved in this [here](https://github.com/rstudio/rticles/pull/419#issuecomment-888295487), but I am open to other suggestions. 

> 
> So I am hesitant on what to do exactly:
> 
>     * Create a new function
> 
>     * Have a switch is this function to use the old or new template ? We would have both and copy / use the correct one depending on what is asked. We could still have the old as default maybe but this could allow us to deprecate the function and inform of changes before changing the default.
> 

I think creating a new function (e.g., `oup_new_format_article`) until such time as almost all OUP journals use the new format is probably the best option. For reasons I explain [here](https://github.com/rstudio/rticles/pull/419#issuecomment-888295487), a switch seems hard to me. A new function would preserve backward compatibility and make it clearer what the options are.

> 
> I am not quite sure of the difference however and it seems this could be minimum.
> 
> > I didn't include the .csl file as they are on CTAN. However, as what is on CTAN doesn't include a .sty file, I am not 100% sure that the style is easily installable with tlmgr. I don't use tinytex, so it is hard for me to check if that is the case or not.
> 
> Do you mean `cls` file ? The new file will be found so we are ok
> 
> ```
> > tinytex::parse_packages(files = "oup-authoring-template.cls")
> tlmgr search --file --global "/oup-authoring-template.cls"
> [1] "oup-authoring-template"
> ```

Yes, I meant CLS (I always mix up those two). I tried what you did a couple of weeks ago and it didn't work, but I most likely used .csl instead of .cls...

> 
> If you meant CSL, where is it available ,
> 
> > This new template is largely backward compatible with the old template, but there was no simple way to maintain full backward compatibility. In particular, multi-paragraph abstracts in the old format won't work in the new.
> 
> I understand that my concerns are to be minimized because you are saying this is backward compatible, is that right ? Do you know the list of changes that we could document / inform ?
> 

The new template is "largely" backward compatible. The main breaking issue I see is multi-paragraph abstracts. The old format used `\begin{abstract}...\end{abstract}`, so multi-paragraph abstracts was easy, whereas the new format used a command `\abstract{paragraph 1\\paragraph 2}`, so the only solution was to make the abstract element of the YAML header a list/vector, each element of which is treated as a paragraph.

The other major change is that now `citation_package` is by default "natbib", whereas before it was "default".

> > I discovered that not all OUP Journals currently accept this new format. In particular, one that I submit to, ICES Journal of Marine Science, does not use this format. As such, I will submit a separate pull request creating a format for this specific journal that is still based on the old OUP style.
> 
> This could mean also that ICES JMS may not be the only one. is `icesjms_article` required if you were able to still use somehow the current `oup_rticles()` function ?
> 

Yes, there are undoubtedly other journals that still use the old format, but I do not have an exhaustive list. If we want to keep backward compatibility, I think that the easiest solution is to keep the old oup_article format as is and create a new function for this new format. 

> Thanks again for you valuable contribution !

**Comment by dmkaplan2000** [comment]
> > I looked into the R-CMD-check errors and they appear to be due to older versions of pandoc not understanding certain elements of my template.tex (specifically, pandoc 2.7.3 does not understand $elseif()$ and pandoc 2.9.2.1 does not seem to know about the pipe "first"). The development version of pandoc also produces an error, but that is not related to the OUP template. The strange thing is that on my system running Ubuntu 20.04 with pandoc 2.5, everything works fine.
> 
> For the CI error, it would require that master should be merged. There are some fixes already for some error in other template. As of now, all templates are tested in a PR and not just the one you change.
> 
> About the pandoc requirement, yes template syntax is dependant of Pandoc version.
> 
>     * Pipes where introduces in Pandoc 2.10 https://pandoc.org/releases.html#pandoc-2.10-2020-06-29
> 
>     * `elseif` in Pandoc 2.8: https://pandoc.org/releases.html#pandoc-2.8-2019-11-22
> 
> 
> It should not work in Pandoc 2.5 for you.

And yet it does... 

```r
> system("pandoc -v")
pandoc 2.5
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.7.7
Default user data directory: /home/dmk/.pandoc
Copyright (C) 2006-2018 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

But I just noticed that RStudio isn't running the default pandoc version, but rather a custom installation that is newer:

```r
> system("/usr/lib/rstudio/bin/pandoc/pandoc -v")
pandoc 2.11.4
Compiled with pandoc-types 1.22, texmath 0.12.1, skylighting 0.10.2,
citeproc 0.3.0.5, ipynb 0.1.0.1
User data directory: /home/dmk/.local/share/pandoc or /home/dmk/.pandoc
Copyright (C) 2006-2021 John MacFarlane. Web:  https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
```

> 
> We could make a minimal requirement on Pandoc for a format. I am open to that. However, it would justify a new format or an opt-in way to not impact user of current `oup_article`

Pipes and ifelse are really pretty useful. I would say that we make a new function with a minimum pandoc version (I am not sure technically how to impose this...).

R/article.R
```diff
@@ -281,19 +281,67 @@ mnras_article <- function(..., keep_tex = TRUE, fig_caption = TRUE) {
...
+  if (onecolumn)
+    pandoc_arg_list = c(pandoc_arg_list,
+                        rmarkdown::pandoc_variable_arg("onecolumn"))
+
+  # section numbering
+  if (!number_sections)
+    pandoc_arg_list = c(pandoc_arg_list,
+                        rmarkdown::pandoc_variable_arg("unnumsec"))
+
+  pandoc_arg_list <- unlist(pandoc_arg_list)
```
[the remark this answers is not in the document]
dmkaplan2000: I didn't previously see these messages for some reason. 

The reason I placed these in the format function is that this way you can quickly change between article formats (for example, if the article is rejected in one journal, you can quickly switch to another). This is also why I have tried to follow the overall skeleton/YAML of Elsevier articles. In general, I tried to place all things that are likely to be specific to the OUP format in the formatting function. You can still, however, put them at the base level, it just makes the Rmd file a bit less portable to another journal. In my opinion, this is how all journal formats should do it and is more consistent with how the base rmarkdown format works (e.g., `pdf_document` has the `citation_package` and `latex_engine` arguments, but `html_document` does not, and these arguments are not base YAML entries). 

One example of why I did this: the oup cls file has specific non-standard definitions of what `papersize` means. If I left it as a base YAML entry, then when you switch to another journal, it will fail or do strange things as that other journal will likely call `a4` what OUP calls `large`.

R/article.R
```diff
@@ -281,19 +281,67 @@ mnras_article <- function(..., keep_tex = TRUE, fig_caption = TRUE) {
...
+  md_extensions = c("-autolink_bare_uris"),
+  journal=NULL,pandoc_args=NULL,
+  number_sections=TRUE,
+  citation_package=c("natbib","default"),
+  papersize=c("large","medium","small"),
+  document_style=c("contemporary","modern","traditional"),
+  namedate=FALSE,onecolumn=FALSE
 ) {
+  citation_package <- match.arg(citation_package)
+  if (length(citation_package) > 1) stop("Please choose just one ", dQuote("citation_package"))
```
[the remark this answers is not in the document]
dmkaplan2000: OK, will change.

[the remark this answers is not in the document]
dmkaplan2000: Note that I got this code from a line in the `ims_article` format. This would presumably also need to be corrected, but I am not going to do it as I don't want to unnecessarily mess around in someone else's code.

R/article.R
```diff
@@ -281,19 +281,67 @@ mnras_article <- function(..., keep_tex = TRUE, fig_caption = TRUE) {
...
+#'   \url{https://www.ctan.org/pkg/oup-authoring-template}.
 #' @export
 #' @rdname article
 oup_article <- function(
   ..., keep_tex = TRUE,
-  md_extensions = c("-autolink_bare_uris")
+  md_extensions = c("-autolink_bare_uris"),
+  journal=NULL,pandoc_args=NULL,
+  number_sections=TRUE,
+  citation_package=c("natbib","default"),
```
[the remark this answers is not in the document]
dmkaplan2000: OK, will do.

R/article.R
```diff
@@ -281,19 +281,67 @@ mnras_article <- function(..., keep_tex = TRUE, fig_caption = TRUE) {
...
+
+  pandoc_arg_list <- unlist(pandoc_arg_list)
+
   pdf_document_format(
-    "oup",
-    keep_tex = keep_tex, md_extensions = md_extensions, ...
+    "oup", keep_tex = keep_tex,
+    md_extensions=md_extensions,
+    citation_package = citation_package,
+    pandoc_args = c(pandoc_arg_list, pandoc_args),number_sections=FALSE,
```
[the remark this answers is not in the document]
dmkaplan2000: The thing is that the OUP cls file has an entirely different mechanism for numbering sections via the `unnumsec` option to the document class. So as to be able to use the standard `number_sections` YAML entry, but not have double and potentially conflicting code in the output, I thought I needed to set a custom `unnumsec` argument to activate this mechanism, but then unset the standard mechanism via the `number_sections` argument to `pdf_document`. But this was because I never went over the default latex pandoc template. I now see where in that document how section numbering comes about and will adapt my template appropriately to take advantage of the existing mechanism.

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -1,8 +1,31 @@
+% Colors for colorlinks
 $if(colorlinks)$
 \PassOptionsToPackage{dvipsnames,svgnames*,x11names*}{xcolor}
 $endif$
 
-\documentclass[$if(fontsize)$$fontsize$,$else$12pt,$endif$$if(spacing)$$spacing$,$else$halfline,$endif$$if(papersize)$$papersize$,$else$a4paper,$endif$$for(classoptions)$$classoptions$$sep$,$endfor$]{ouparticle}
+% Set document class options
+\documentclass[webpdf$if(papersize)$,$papersize$$else$,large$endif$$if(document_style)$,$document_style$$else$,contemporary$endif$$if(unnumsec)$,unnumsec$endif$$if(namedate)$,namedate$endif$$if(classoptions)$,$for(classoptions)$$classoptions$$sep$,$endfor$$endif$]{oup-authoring-template}
```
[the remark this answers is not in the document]
dmkaplan2000: This has now been dealt with in comments elsewhere, but to be clear, a document made with the old skeleton will look different in the new format, but have the same content. There are a few breaking changes I have mentioned elsewhere, but this is taken care of by the versioning scheme we have decided on elsewhere.

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -1,8 +1,31 @@
...
+
+% one column
+$if(onecolumn)$\onecolumn$endif$
+
+%\usepackage{showframe}
+
+% line numbers
+$if(numberlines)$\usepackage{lineno}\linenumbers$endif$
```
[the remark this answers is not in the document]
dmkaplan2000: Good question. I threw this code in the template in case one day we wanted to use it, but never made reference to it in the Rmd template as I was unsatisfied with it and preferred to suggest that as a `header-includes`. If we were to do it, then I think the correct way to go about it so as to be consistent with `number_sections` would be to include it as a `number_lines` variable in the formatting function that is translated in `numberlines` pandoc arg in the function. We would also need a `number_lines_options`:

```
output:
  rticles::oup_article:
    number_lines: TRUE
    number_lines_options: ["mathlines","switch"]
```

Alternatively, we could make the `number_lines` argument do double duty, but that might be confusing:

```
output:
  rticles::oup_article:
    number_lines: ["mathlines","switch"]
```

In either case, I am not sure how to convert a vector of values in the formatting function into variables passed to pandoc (I do this using `rmarkdown::pandoc_variable_arg` for single value variables, but not sure what the appropriate syntax is for multiple value variables.

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -1,8 +1,31 @@
...
+\newtheorem{theorem}{Theorem}%  meant for continuous numbers
+%%\newtheorem{theorem}{Theorem}[section]% meant for sectionwise numbers
+%% optional argument [theorem] produces theorem numbering sequence instead of independent numbers for Proposition
+\newtheorem{proposition}[theorem]{Proposition}%
+%%\newtheorem{proposition}{Proposition}% to get separate numbers for theorem and proposition etc.
+\theoremstyle{thmstyletwo}%
+\newtheorem{example}{Example}%
+\newtheorem{remark}{Remark}%
+\theoremstyle{thmstylethree}%
+\newtheorem{definition}{Definition}
```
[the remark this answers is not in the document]
dmkaplan2000: This is just boiler plate code from the standard OUP template that I personally would not use and so I wasn't sure how to best make it accessible to the user. It seems like this could go in `header-includes`, though it is quite a bit of code.

inst/rmarkdown/templates/oup/resources/template.tex
```diff
@@ -1,8 +1,31 @@
...
 
 % Packages I think are necessary for basic Rmarkdown functionality
```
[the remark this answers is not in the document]
dmkaplan2000: This list is largely because I built my original OUP template off of another existing rticles template (can't remember which) without getting to know the default pandoc latex template. Now that I know that, I see that there are potentially improvements to be made, but am not too enthusiastic about making them as everything is working. If I see easy cleanup opportunities, I will do it, otherwise, I will leave it.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: Yes. I copied this from somewhere before understanding how `$highlighting-macros$` works.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: This is my code to make their template format work with author affiliations and footnotes that are specified by ID code in the Rmd skeleton. This was horrifyingly difficult to figure out, but I can't see a better way and now the code is sufficiently generic that I think it might be useful elsewhere in other rticles formats.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: I have tried to make this not be a breaking change for articles written using the old format. The code you highlight has two functions:

1) The first part associates codes with addresses. This is necessary for dealing with multiple authors with the same address.
2) Format addresses that are given in the new format that is split into a bunch of pieces.

If users used the older `$address.address$`, then it should still work.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: My understanding is that this is customized by the journal itself in post-processing after paper acceptance. This is why I didn't attempt to make these options for the YAML header.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: Again, my understanding is that this would be added in post-processing by the journal itself.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: I think I did this as other formats did this and these sections seemed special and I was worried that the CLS might do something special with these sections. It is also information that may be excluded for double-blind reviewing, so somewhat convenient to have it all in the YAML header.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: The problem is that the document class has an option called "namedate" that is separate from the bibliographystyle, which is also given in the default template. I don't fully understand why they did it this way, separating one piece of information into two parts, and I fully agree that it is suboptimal, but that is how it is...

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: This has been moved into the CLS file as a required package. 

```latex
\RequirePackage{natbib}
```

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: The relevant code from the CLS file is:

```latex
\DeclareOption{namedate}{\PassOptionsToPackage{authoryear,round}{natbib}}
```

But in the OUP template it has:

```latex
%USE THE BELOW OPTIONS IN CASE YOU NEED AUTHOR YEAR FORMAT.
%\bibliographystyle{abbrvnat}
```

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
dmkaplan2000: It is very format specific. I want to make it as effortless as possible to change from one journal format to another.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
dmkaplan2000: I explain above why I chose `header-includes` instead of showing the option. But I am open to rethinking things provided a good way of passing options to `$number_lines$` is found.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
dmkaplan2000: These are already in the template and skeleton:

* `unnumsec` => `number_sections` YAML entry
* `contemporary`, `modern`, etc. => `document_style` YAML entry
* `large`, `medium`, etc. => `papersize` YAML entry

I haven't done anything with the `webpdf` option as that is always given in all the examples they show and in reality I am not 100% sure what it does.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: Thanks for the suggestion. Done.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
dmkaplan2000: Removed the comment.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: Now fixed.

R/article.R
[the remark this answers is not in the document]
dmkaplan2000: I think I misspoke here. The new format will not work with `biblatex`, which is why I use match.arg on it and limit it to these two options. I will fix shortly to return the code to its original format.

R/article.R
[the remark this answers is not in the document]
cderv: That are good points. 

Currently **rticles** formats and template are not designed with switching the format function in the template. That is why the base level variable in metadata are used a lot. This is also because it is easy and avoid having a logic inside a format to pass them as variable. The command line will be rather long. but we can live with that. 

So I understand the idea. I'll think about it because maybe it is a better way to deal will all the templates. 

Currently we are more trying to reach for an uniformisation of the variables names and same kind of YAML header for all template. 

> and is more consistent with how the base rmarkdown format works (e.g., pdf_document has the citation_package and latex_engine arguments, but html_document does not, and these arguments are not base YAML entries).

About this, we do that mainly because we need to know the value for some processing, and we can even modified it or pass other option. So basically, they are not passed directly as pandoc variables. For any variable that can be passed to Pandoc directly as Pandoc variables without any more logic, then the usual metadata variable syntax is used.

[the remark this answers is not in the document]
dmkaplan2000: Yes, I get that `citation_package` et al. have a somewhat different role than `papersize`, but conceptually I think it makes more sense to put format specific options in the format function. If it is not hard to put the options in the format function, then I don't see why not to do it. This all goes along with my "generic article format" I proposed elsewhere. The idea would be to write the article once using the generic format and then just change the format function to get to that of a specific journal. This seems like a good long-term goal for rticles, though I understand this will require quite a bit of work.

R/article.R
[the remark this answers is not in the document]
cderv: It is now done f171c8f0

thanks for the pointer.

R/article.R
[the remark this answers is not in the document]
cderv: OK it was my understanding and wanted to confirm. It would be good to add a note in the help page for this function about biblatex not being supported.

R/article.R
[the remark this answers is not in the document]
cderv: I don't mind setting the `number_section = FALSE`. However, if we do so we should not expose the argument in the format function. Just hard code the FALSE to prevent user from changing.

inst/rmarkdown/templates/oup/resources/template.tex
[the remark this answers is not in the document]
cderv: >In either case, I am not sure how to convert a vector of values in the formatting function into variables passed to pandoc (I do this using rmarkdown::pandoc_variable_arg for single value variables, but not sure what the appropriate syntax is for multiple value variables.

I don't think you can pass such values using command line `-V` flag. From doc, https://pandoc.org/MANUAL.html, this would require using a [metadata file ](https://pandoc.org/MANUAL.html#option--metadata-file). So it gets more complicated on the code base part.

And that is also why we are trying to pass most option as base level of YAML header because Pandoc knows how to parse this YAML header and can handle better much complex variable. 

This would be

````
numberlines: TRUE
numberlines-options: ["mathlines","switch"]
````

and the template would work I guess.

inst/rmarkdown/templates/oup/resources/template.tex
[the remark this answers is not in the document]
cderv: `header-includes` from pandoc requires text but `pdf_document()` has a `includes` variable, with a `in_header` value that takes a file. So you could add such definition in a `preamble` tex file and pass it to the format

````
output: 
  oup_article:
    includes:
      in_header: preamble.tex
````

inst/rmarkdown/templates/oup/resources/template.tex
cderv: We could then benefit to have that in an external file and make it available in different templates ? Is that what you are saying ?

inst/rmarkdown/templates/oup/resources/template.tex
cderv: Ok thanks !

inst/rmarkdown/templates/oup/resources/template.tex
cderv: OK ! thanks!

inst/rmarkdown/templates/oup/resources/template.tex
cderv: >  think I did this as other formats did this and these sections seemed special and I was worried that the CLS might do something special with these sections

Some templates wants this in places that are not adapted to being at the end of the body. Hence the YAML trick to make them appear in specific place in the template, outside of body. 

> It is also information that may be excluded for double-blind reviewing, so somewhat convenient to have it all in the YAML header.

This is ok to me. Let's keep it that way !

inst/rmarkdown/templates/oup/resources/template.tex
cderv: if we are ok to no allow user to set another bibliostyle than these two, then we should keep. Otherwise, we need to find a solution. 

This could be documentation only like 

````
# biblio-style: abbrvnat # use this if you need AUTHOR YEAR FORMAT (oup option `namedate`)
````

inst/rmarkdown/templates/oup/resources/template.tex
cderv: Oh ok ! 

So `natbib` is really the only possible citation package that you can use. You could not use `pandoc` citation at all with this template ? Is that right ? 
Because even if you don't want to use natbib (`citation_package: default`), the package `natbib` will be required. (and maybe could lead to issues ?)

If so, we need to force `natbib` in the format function then.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
cderv: I understand. we need to discuss that. We have the other thread.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
cderv: I am just saying here that I think we need one of the other. 

I am fine with showing what to pass in header-includes, but in that case I think we should not add the line in the template. 

It is the duplication that bothers me.

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
cderv: Oh ok. I missed that. I need to recheck how this works.

R/article.R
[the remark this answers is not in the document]
dmkaplan2000: I already pushed something that fixes this issue and completely eliminates the confusion. I think this is the best solution.

inst/rmarkdown/templates/oup/resources/template.tex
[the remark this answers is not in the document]
dmkaplan2000: Can we meet half way and underlying in the template I use numberlines and numberlines-options, but in the format function also create the possibility of using `number_lines` and `number_lines_options: "mathlines,switch"`? I really like keeping format specific stuff in the format function (line numbering is, admittedly, pretty generic, but it wouldn't make sense, for example, in a HTML document...).

inst/rmarkdown/templates/oup/resources/template.tex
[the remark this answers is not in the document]
dmkaplan2000: I opted to drop it in the `header-includes`. It won't win an award for beauty, but it works. If it is an issue, I can move it out to a file as you suggest.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: Could be. We would have to look to see how other formats and style files deal with multiple author affiliations and author notes. In my case, I had to build this as it isn't in the OUP CLS file itself, but I know that this sort of stuff is built into the Elsevier CLS files, avoiding the need for this sort of gymnastics.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: The OUP template says these two formats should be used. I haven't tried others to see what will happen though. I am OK with just allowing these two options.

inst/rmarkdown/templates/oup/resources/template.tex
dmkaplan2000: Well... With my template, even if `natbib` is a required package, pandoc citations using `.csl` files will work just fine. I thought initially that I would force the user to use `natbib`, but if `.csl` citations can also be made to work, why not. I think one of the reasons that more OUP journals have not adapted the new format is specific citation styles, so if we can use pandoc to produce those formats, then perhaps the journals will be happy to accept the articles...

R/article.R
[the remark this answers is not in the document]
dmkaplan2000: I have cleared this up in the documentation of the format function on the `newoldouptemplate` branch associated with PR #431.

Une chose rien à voir: Vous n'habitez pas à Montpellier par hasard ?

inst/rmarkdown/templates/oup/skeleton/skeleton.Rmd
dmkaplan2000: I have implemented the "format function" solution on the branch linked to PR #431. I did this in two steps, first implementing it at the base YAML level using `numberlines` and then moving it into the format function in commit `5f143751272242eb975f70cf8adf299f67fbd2ab`. If ever it is really a problem to put this option into the format function, then we can just undo this single commit to get back to the base level configuration.

dmkaplan2000:
I have also imposed a minimum pandoc version for oup_version>0 in the branch associated with PR #431.

cderv:
Should be superseded by #431 right ?

dmkaplan2000:
> Should be superseded by #431 right ?

I believe this is the case, but it has been a while since I thought about it.

cderv:
From the content of the PR, the updated is now available in `oup_article(oup_version = 1)`. So I believe this should cover what this issue was about. 

I'll close this, and if some newer update are needed, we'll do it in the new version. Thanks!

dmkaplan2000:
No problem, thanks...

Issue #182, opened 2018-08-03
pdf knit fails with 'Environment Shaded undefined' Error for IEEE template --> stackoverflow solution


rainer-rq-koelle:
Dear all,
((I am not sure whether this is a feature request. But a bug report, did not feel right.))
I struggled heavily to work through a sort of trivial problem (c.f. [my question](https://stackoverflow.com/questions/51594999/unable-to-make-sense-out-of-rstudio-pdf-knit-fails-with-environment-shaded-und) and [a related question](https://stackoverflow.com/questions/41052687/rstudio-pdf-knit-fails-with-environment-shaded-undefined-error)):

Using the rticles IEEE paper template, I wanted to **print a R code-chunk**. 

The default workaround offered on Stackoverflow is to "circumnavigate" the problem by not showing the chunk (aka setting the chunk option `echo = FALSE` which is not the desired behaviour).
It took me a while to figure out where, i.e. `template.tex` in the respective rticles R library sub-folder to change the LATEX for a possible solution.

The solution is to add a 3 line LATEX statement in the preamble of `template.tex` to support the highlighting: 
  ```
  $if(highlighting-macros)$
    $highlighting-macros$
    $endif$ 
```

_Mickey_ provided an elegant solution in response to the aforementioned fix identified by _coatless_ that is based on a separate `header.tex` file. I know too little about LATEX, but given the number of lines I find this solution pretty sophisticated.

Still, I believe, the support for the highlighting, i.e. Shaded environment, should be included in the rticles templates **out of the box**.

Thanks for considering the fix offered by _coatless_ in the next update of the IEEE paper (and possible other templates).

yihui:
Many thanks for asking and answering the question on SO! I have added the fix to `ieee_article()`.

rainer-rq-koelle:
Great, thanks, Yihui.

m-macaskill:
Under rticles version 0.9, this error still occurs for PeerJ-formatted documents.

yihui:
@rainer-rq-koelle Cc @zkamvar (#127)

yihui:
@m-macaskill Fixed now by @zkamvar via #238. Thanks!

Issue #99, opened 2016-12-09
! LaTeX Error: Environment Shaded undefined.


HAVB:
When trying to knit a simple test document using the ACM format, I get the following error:

```
! LaTeX Error: Environment Shaded undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.45 \begin{Shaded}

pandoc: Error producing PDF
Error: pandoc document conversion failed with error 43
Execution halted
```

I can knit the same document as regular PDF with no problem.

HAVB:
After further examination, this may not be a bug after all. There's no error when I hide the code chunks with `echo = FALSE`. I guess that for an ACM style paper you are not supposed to show the code behind your plots and other output, and that's why functionality for code highlighting is not implemented.

See http://stackoverflow.com/questions/41052687/rstudio-pdf-knit-fails-with-environment-shaded-undefined-error  for details

PR #422, opened 2021-07-20
Add highlighting macros for Pandoc in JSS template

vincentarelbundock:
Compiling a JSS article results in an error about "Shaded Environment undefined". The same error was previously reported for different article types here:

https://github.com/rstudio/rticles/issues/182
https://github.com/rstudio/rticles/issues/99

The solution is to add this to the template:

```latex
$if(highlighting-macros)$
$highlighting-macros$
$endif$
```


cderv:
Hi @vincentarelbundock, 

Can you just share a reproducible example with the issue ? 

I want to understand what we missed for not having an error in our test about that. 

With the default template, I don't have an issue and it contains a R chunk. So I want to be sure we got it correctly covered. 

I assumes this is for other language chunk ? 

The fix is the correct one, don't get me wrong. 

Thanks for the PR !

vincentarelbundock:
I just tried to produce a MWE, but for some reason I can no longer replicate the problem.

Sorry!

There are so many parts interacting here. It's a fairly complicated article with `kableExtra` tables, colors, and tikz diagrams. I also vaguely remember having more issues on some of my machines than others (I run a mac laptop, windows machine, and linux in WSL2).

Would love to help more, but can't at the moment.

Feel free to ignore the PR if you need a test or example.

cderv:
In your article, 

* do you have some non R chunk ? 
* do you have some verbatim code (non executed) that you apply a class language too ? like
  ````markdown
  ```yaml
  title: hello
  ```
  ````

This is the two case I need to reproduce.

vincentarelbundock:
sorry, should have answered that. No, it's only R, and no verbatim environment with class language. There are some `eval=FALSE`, but that's it.

cderv:
Ok thanks. I'll see what I'll do. If you got the error again, please come back and share! 

Thanks again.

nbenn:
I might be able to contribute an example that triggers the issue @vincentarelbundock is trying to address. Creating the JSS draft .Rmd using `rmarkdown::draft()` and adding a chunk as

    ```r
    foo <- "bar"
    ```

for me (macOS 11.5, R 4.1.1, rticles 0.20), gives `! LaTeX Error: Environment Shaded undefined.` The reason for this is that the resulting code block is wrapped in `Shaded` and `Highlighting` environments instead of the standard `CodeChunk` and `CodeInput`, one gets when having a chunk like

    ```{r}
    foo <- "bar"
    ```

The proposed PR does resolve the issue such that both ways of specifying a code chunk compile just fine. Visually, however, they are distinct: `Shaded`/`Highlighting` actually provides syntax highlighting, whereas `CodeChunk`/`CodeInput` for JSS is strictly b/w.

cderv:
Thanks. That confirms what I thought. 

As context, let me give some details. 

This syntax
````
```r
foo <- "bar"
```
````
is a Pandoc direct syntax. It will apply class `r` to the code so that Pandoc's highlighting gets triggered for R syntax. Hence the highlighting. This is why the LaTeX error happens because we need the environment definition required by Pandoc's Highlighting. Using this syntax won't evaluate the code. 

However, using 
````
```{r}
foo <- "bar"
```
````
will evaluate the code and we specifically output with `CodeChunk` and `CodeInput` (there is `CodeOutput` too) which are defined in `jss.cls`. They don't highlight the same obviously.

So we can add the missing definition for Pandoc's Syntax Highlighting. However, I don't how correct it is for JSS submission if a document contains these environments.

nbenn:
Thanks @cderv!

> I don't how correct it is for JSS submission if a document contains these environments.

It's probably not. However JSS papers are not the only use-case for `rticles::jss_article`. I was using this in a package vignette with the `nojss` classopt.

In my post, I was not trying to say, this should or should not be merged. That's up to you @cderv. I simply commented b/c this issue caused me a bit of a headache to track down. For some reason, the rhub machines and CRANs r-devel-windows-ix86+x86_64 machine did not trip over this, only r-devel-linux-x86_64-debian-gcc. Just out of curiosity, @cderv, does that make sense to you that this works in some cases and does not in others?

I guess, with my troubles in mind, I'd be slightly in favor of merging this. Just to make people's life easier. But that's just my opinion.

cderv:
Thanks @vincentarelbundock !

vincentarelbundock:
Thank you for all your work on this great package!

PR #423, opened 2021-07-20
changes reference location in Frontiers template

graysonwhite:
Hi there, love the package. I have been using it to typeset an article for Frontiers, and found a solution to moving the references to under the "References" header in Rmarkdown rather than at the end of the document. The solution was found from Andrew Morris' answer to [this stackoverflow](https://stackoverflow.com/questions/28815781/how-to-move-the-bibliography-in-markdown-pandoc) question. Hope this helps!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rticles
 Type: Package
 Title: Article Formats for R Markdown
-Version: 0.20.2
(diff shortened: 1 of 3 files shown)


cderv:
Thank you ! 

This will indeed work well for this format which uses Pandoc citeproc only !

PR #431, opened 2021-07-29
oup_article() support for old and new template

dmkaplan2000:
Update to OUP format including both old and new templates, with a parameter in the formatting function controlling switching between the two.

It is generally working, but the testing part of the package will not work because it currently assumes that 1 format = 1 directory with the same name, which is not the case for this format. I will leave it to you to correct this.

----------------------------------------------------------------

- [ X] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [ X] Add a skeleton article `inst/rmarkdown/templates/journalname/skeleton/skeleton.Rmd`.

- [ X] Add a description of the template `inst/rmarkdown/templates/journalname/template.yaml`.

- [X ] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. If you are using TinyTeX or TeX Live, you can verify if a package is available on CTAN via `tinytex::parse_packages(files = "FILENAME"")` (e.g., when `FILENAME` is `plain.bst`, it should return `"bibtex"`, which means this file is from a standard CTAN package). Please keep the number of new files absolutely minimal (e.g., do not include PDF output files), and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include too many items in it without using all of them).

- [X ] Update Rd and namespace (could be done by `devtools::document()`).

- [X ] Update NEWS.

- [X ] Update README with a link to the newly supported journal. Please add your Github username and the full name of the journal (follow other examples in the list). 

- [--- ] Add a test to `tests/testit/test-formats.R` by adding a line `test_format("journalname")`. We try to keep them in alphabetical order.

- [X ] Add your name to the list of authors `Authors@R` in DESCRIPTION. You don't need to bump the package version in DESCRIPTION.

Thank you!


cderv (approved):
Hi @dmkaplan2000 !

I have finished the PR based on our previous discussions. Thanks a lot for the hard work on this and all the answers you gave. 

We have now one `oup_article()` but using two templates based on `oup_version()`. 
Hopefully this should solve the issue and suits the need of #418 and #419, right ? 

About the CLA, I think you have already send it from your previous work. 

I'll merge this PR and we can adapt for the other needs if somethings are missing.

dmkaplan2000:
Thanks for this. I will look at the new version down the road, but I am a bit busy right now, so it won't be right away.

PR #434, opened 2021-08-16
Allow classoptions

statzhero:
Such as `letterpaper`


cderv (approved):
Thank you ! 

I just modified to support multiple classoptions as other templates. Hope that is ok!

PR #251, opened 2019-10-16
Add Journal of Educational Data Mining

jooyoungseo:
# Changes

- Added Journal of Educational Data Mining LaTeX template according to the code of conduct.
- Added a `jedm_article()` function in `article.R`.
- Added related info to `README.md`; `DESCRIPTION`; and `NEWS.md`.

# Building

- Used `Rd2roxygen::rab()` function to build `rticles` package after change has been made.

# Testing

I have confirmed that `rticles::jedm_article()` output format works.

# Signed-Form

- I have signed and submitted the RStudio form to J.J.

PR #438, opened 2021-09-07
Updated JooYoung's info

jooyoungseo:
Thanks for your final touch, @cderv! This PR updates my information (e.g., email; affiliation) following #251.


cderv (changes requested):
Thanks! 

I agree that DESCRIPTION should be updated. However, maybe we should not put your information into the skeleton Rmd so that your information are not left in all template document created and we do not have to update them while they change. 

Don't you think ?

jooyoungseo:
Agreed, @cderv! I will update the default skeleton author info following the original template by tomorrow.

cderv (approved):
Thanks !

Issue #394, opened 2021-04-30
R file produced by `rjournal_article` is not replaced by new one


cderv:
We forgot to set `overwrite = FALSE` here I think: 
https://github.com/rstudio/rticles/blob/84a8fb4d705990cf455ff8cb1e0c8340d04bda0a/R/rjournal_article.R#L94

Meaning that the new purled R file was not copied to the output dir unless the previous one was cleaned. No error or warning issued with `file.copy` in this case... 

That is nasty bug I think. But with an easy fix

Thanks @apreshill for finding this

Enchufa2:
I suppose this is a feature to automatically generate the replication script from the Rmd. But I always maintain a separate `filename.R` replication script, and now it gets overwritten after this change. Is there an option to switch off this feature?

Enchufa2:
I see no way to switch this off. And given that previously it was possible to maintain a custom replication script, I consider this a regression. A proper fix would be to add some comment to the generated file (like "this was automatically generated, do not edit by hand"), and only overwrite it if such comment is present.

cderv:
Hi @Enchufa2 ! 

Thanks for the report. This is indeed something we did not think through.

We could also provide an argument in the function to opt-out this: 
````yaml
output: 
  rjournal_article: 
    purl: false
```

I like your suggestion though with the special comment in the script. 

I'll open an issue to track your report specifically.

Enchufa2:
Nice, thanks!

PR #439, opened 2021-09-09
Only overwrite the purled file if this our file.

### LINKED ISSUE #433: Do not overwrite any user own existing replication R script for rjournal_article
This was originally posted in https://github.com/rstudio/rticles/issues/394#issuecomment-894316281

We need to watch out to not override any existing R file that would be owned by the user and not generated by us. 

Two ideas: 

* Provide an opt-out mechanism for the purl-ing. 
  Something like 
  ````yaml
  output: 
    rjournal_article: 
      purl: false
  ````

* Use a special comment header `# Do not edit` as done in other tools so that we know that this is our file and we can overwrite.

cderv:
This closes #433 by providing a new mechanism into `rjournal_article()` to avoid overwriting any existing R file user maintained

# Current behavior

* `rjournal_article()` will purl the Rmd file to produce an R file with the same name as the output name for tex file and pdf file. 
* Since version 0.20, we were overwriting any existing R file. 
* This causes issue for users who maintain their own R script and don't want the purled one.

See https://github.com/rstudio/rticles/issues/394#issuecomment-894316281

# New behavior

* `rjournal_article()` generates a file which add a special custom header comment so that RStudio IDE sees this file as read only
* With this header we can detect if file is our own or not. 
* If we see the R file is not generated by **rticles**, a warning is issued and purled file is no copied into the output folder. 

# Possible other solution

* We do less magic and just offer a switch in the format to deactivate the purling
  ```yaml
  output: 
    rjournal_article: 
      purl: false
  ```
  which would default to true, but can be set to false for those who needs to maintain their R file. 

@yihui can you review this to give your thoughts on the current solution and possible other one ? 

Thank you.


yihui (approved):
Looks clever and good to me! Thanks!

PR #446, opened 2021-11-09
Update Copernicus to v6.4 and add missing Copernicus journal

RLumSK:
Another update of the Copernicus template (now version 6.4) and I've added a so far missing journal ("Web Ecology") to `copernicus_article()`

- [x] Update NEWS.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # rticles 0.22
 
-- Update Copernicus Publications template to version 6.4 from 2021-08-16
+- Update Copernicus Publications template to version 6.4 from 2021-08-16 (thanks, @RLumSK, #446).
  
 # rticles 0.21
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # rticles 0.22
 
-
+- Update Copernicus Publications template to version 6.4 from 2021-08-16
+ 
 # rticles 0.21
 
 ## NEW FEATURES
--- a/R/copernicus_article.R
+++ b/R/copernicus_article.R
@@ -13,7 +13,7 @@
 #'
 #' An number of required and optional manuscript sections, e.g. `acknowledgements`, `competinginterests`, or `authorcontribution`, must be declared using the respective properties of the R Markdown header - see skeleton file.
 #'
-#' **Version:** Based on `copernicus_package.zip` in the version 6.3, 8 July 2021, using `copernicus.cls` in version 9.32, 7 July 2021.
+#' **Version:** Based on `copernicus_package.zip` in the version 6.4, 8th August 2021, using `copernicus.cls` in version 9.32, 7 July 2021.
 #'
 #' **Copernicus journal abbreviations:** You can use the function `copernicus_journal_abbreviations()` to get the journal abbreviation for all journals supported by the Copernicus article template.
 #'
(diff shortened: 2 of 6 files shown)


RLumSK:
@nuest Sorry, I forgot to ping you for this PR.

RLumSK:
@cderv The failed CI checks do not seem to be related to my PR.

cderv:
Yes I agree. It seems to be from the Glossa format. 

our test suites runs every format for now. It would be interesting to only test the modified format for PR 🤔

RLumSK:
@cderv Thanks for the prompt response. For what it is worth, I have tested the modified version, only the Copernicus template, manually on my computer (macOS 11.6.1, pandoc 2.16.1, R 4.1.1). No problem here, but perhaps this is not a replacement for the full tests you run.

R/copernicus_article.R
```diff
@@ -77,18 +77,17 @@ copernicus_journals <- list(
...
   "Archives Animal Breeding" = "aab",
-  "ASTRA Proceedings" = "ap",
   "Atmospheric Chemistry and Physics" = "acp",
   "Atmospheric Measurement Techniques" = "amt",
   "Biogeosciences" = "bg",
   "Climate of the Past" = "cp",
   "DEUQUA Special Publications" = "deuquasp",
   "Drinking Water Engineering and Science" = "dwes",
-  "European Journal of Mineralogy" = "ejm",
   "Earth Surface Dynamics" = "esurf",
```
cderv, line 86: Do we really need to deleted them ? 

This will probably break some code if one was using for example 
````r
rticles::copernicus_journal_abbreviations(journal_name = '"ASTRA Proceedings')
````

Can't we just keep them ? 

possibly in a separate list if we want to differentiate new and old journal name ?

File: NEWS.md [hunk 2]
```diff
@@ -1,6 +1,7 @@
 # rticles 0.22
 
-
+- Update Copernicus Publications template to version 6.4 from 2021-08-16
```
cderv, line 3:
```suggestion
- Update Copernicus Publications template to version 6.4 from 2021-08-16 (thanks, @RLumSK, #446).
```

cderv (approved):
Looks good overall apart from a comment I left above. 

After you answer and possibly change, we are good to merge

R/copernicus_article.R
cderv: > If the journal is not known, the PDF still builds (I've have tested it with rticles + plain LaTeX). 

This is if the function `rticles::copernicus_journal_abbreviations` was not used to derive the journal abbreviation like in the template YAML header, right ? 

> If a journal is not published anymore by Copernicus, so it is unlikely that new articles are written for such journals.

Good point. 

But I like your warning solution to possibly warn there is something not expected. 

> I intended to keep the list precisely as Copernicus has it (easier to compare, and we do not drag old stuff with us)

What about keeping a list of old journal in a separate object and use this list to warn that this is an old journal ? 

What do you think ?

RLumSK: > This is if the function rticles::copernicus_journal_abbreviations was not used to derive the journal abbreviation like in the template YAML header, right ?

No, I used the function, e.g., `r rticles::copernicus_journal_abbreviations(journal_name = 'gaga')` returns 
just something that is empty. More precise, the corresponding code returns `NULL`, which is then included in the template. 

> What about keeping a list of old journal in a separate object and use this list to warn that this is an old journal ?

There is just one problem. I would only know about the list now, and I have no idea what Copernicus did in the past, and I am not sure whether there is a record. Hence, I would prefer to keep that very general. What do you think?

cderv: > More precise, the corresponding code returns NULL, which is then included in the template.

And what happens in that case when no journal is passed ? Just want to be sure to understand everything. THanks

RLumSK: ## Three scenarios: 

1. `rticles::copernicus_journal_abbreviations(journal_name = "")`: All journal abbreviations are passed onto LaTeX; looks odd but the PDF builds.
2. `rticles::copernicus_journal_abbreviations()`: Same as 1.
3. `rticles::copernicus_journal_abbreviations("non-existing journal")`: returns `NULL` and in the LaTeX nothing. 

### Scenario 1 and 2: Tex output

```tex
\documentclass[adgeo, ars, asr, ascmo, angeo, aab, acp, amt, bg, cp,
deuquasp, dwes, esurf, esd, essd, egqsj, ejm, fr, gchron, gh, gc, gi,
gmd, hgss, hess, jbji, jm, jsss, mr, ms, nhess, npg, os, pb, piahs,
sand, sd, soil, se, tc, we, wcd, we, wes, manuscript]{copernicus}
```

### Scenario 3: Tex output

```tex
\documentclass[, manuscript]{copernicus}
```

Bottom line, it is rather robust.

cderv: Ok let's leave it that way then. Thanks for the explanation and the PR

RLumSK:
@cderv Thanks for your support and the critical PR review! You raised an important point I had overlooked, at least now we have tested that and I can keep an eye on it, in case it causes some problems in the future.

Issue #411, opened 2021-06-09
unify YAML variables used accross template by documenting them.


cderv:
Currently, I don't think all the templates use the same Pandoc variable name for a same purposed variable. 

Example: 

- acs template
````
title:
  - short: short title
  - formatted: full title
```

whereas common pandoc is to use 
```yaml
title: full title
shorttitle: short title
```

This thread is to compile the used one and decide which to keep

## List of common variables that can be uses in a template

* `title`: Title of the article
* `shorttitle`: A short title for the article
* `author`: Field for author informations
* `keywords`: Keyword

mingsu:
institute: institute(s) of each author
journalname: journal name
corresponding author(s): including author name name and address.

gregmacfarlane:
`runninghead`: what gets printed at the top of non-title pages. [Wikipedia](https://en.wikipedia.org/wiki/Page_header) suggests "running head" over "running header" or "running title"

PR #461, opened 2022-01-03
Add ISBA template (Bayesian Analysis)

dmi3kno:

- [ X] This project uses a Contributor Licence Agreement (CLA) that you'll be asked to sign when opening a PR. This is required for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). We use a tool called CLA assistant for that.  

- [X] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [X] Document your function using **roxygen2**. Markdown syntax is supported. Refer to <https://roxygen2.r-lib.org/articles/rd-formatting.html> for formatting references.

- [X] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname/resources/template.tex`.

- [X] Add a skeleton article `inst/rmarkdown/templates/journalname/skeleton/skeleton.Rmd`.

- [X] Add a description of the template `inst/rmarkdown/templates/journalname/template.yaml`.

- [X] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. If you are using TinyTeX or TeX Live, you can verify if a package is available on CTAN via `tinytex::parse_packages(files = "FILENAME"")` (e.g., when `FILENAME` is `plain.bst`, it should return `"bibtex"`, which means this file is from a standard CTAN package). Please keep the number of new files absolutely minimal (e.g., do not include PDF output files), and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include too many items in it without using all of them).

- [X] Update NEWS.

- [X] Update README with a link to the newly supported journal. Please add your Github username and the full name of the journal (follow other examples in the list). 

- [X] Add a test to `tests/testit/test-formats.R` by adding a line `test_format("journalname")`. We try to keep them in alphabetical order.

Thank you!


inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+$header-includes$
+$endfor$
+
+$preamble$
+
+\begin{document}
+\begin{frontmatter}
+
+\title{$title$}
+\runtitle{$title$}
```
cderv, line 96: Could it be different than the title ? 
Some templates are using `runtitle` variable, or `shorttitle`

It could be 
```suggestion
\runtitle{$if(runtitle)$$runtitle$$else$$title$$endif$}
```

We still get a hard time unifying those variable (#411) - this is on the TODO

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+
+\begin{aug}
+$for(author/pairs)$
```
cderv, line 99: This will introduce a Pandoc requirement that needs to be put as a check in the format function, the doc and the test. 
It would be Pandoc 2.10 I believe considering the pipes function used. 


Like in 
https://github.com/rstudio/rticles/blob/9cfda82cb41ecc4706b4fd7f0925d2db7198ce2e/R/article.R#L251-L253

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+\end{aug}
+
+\begin{abstract}
+$abstract$
+\end{abstract}
+
+% MSC class
+$if(MSC2020primary)$ \begin{keyword}[class=MSC]
+$for(MSC2020primary/first)$\kwd[Primary ]{$it$}$endfor$
+$if(MSC2020primary/rest)$$for(MSC2020primary/rest)$\kwd{$it$}$sep $$endfor$$endif$
```
cderv, line 128: I think using only `for` should work - I don't think the `if` is needed here, is it ? 
Like if there is no `MSC2020primary/rest`, then it will do nothing

This would simplify the template

```suggestion
$for(MSC2020primary/rest)$\kwd{$it$}$sep $$endfor$
```

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+$for(include-before)$
+$include-before$
+$endfor$
+
+$body$
+
+$for(include-after)$
+$include-after$
+$endfor$
+
```
cderv, line 144: Usually this goes at the very end of the template so that one can add something after everything the template is doing. 

Currently it will have the same effect to put in `include-after` than at the end of the Rmd file.

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+$if(MSC2020primary/rest)$$for(MSC2020primary/rest)$\kwd{$it$}$sep $$endfor$$endif$
+$if(MSC2020secondary)$$for(MSC2020secondary/first)$\kwd[; secondary ]{$it$}$endfor$
+$if(MSC2020secondary/rest)$$for(MSC2020secondary/rest)$\kwd{$it$}$sep $$endfor$$endif$$endif$\end{keyword}$endif$
+% keywords
+$if(keywords)$ \begin{keyword} $for(keywords)$\kwd{$keywords$}$sep$ $endfor$\end{keyword}$endif$
+\end{frontmatter}
+
+$for(include-before)$
+$include-before$
+$endfor$
```
cderv, line 137: Usually this goes at the somewhere after `\begin{document}` to insert content before some other content where that makes sense. 

Currently it will have the same effect to put in include-before than at the start of the Rmd file.

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+$endfor$$endif$
+
+$if(acknowledgements)$
+\begin{acks}[Acknowledgments]
+$acknowledgements$
+\end{acks}
+$endif$
+
+\bibliographystyle{ba}
+\bibliography{$bibliography$}
```
cderv, line 160: Seems like only `natbib` is supported as those command are not behind a `if`. Is it the case ? 

Can you use with Pandoc citeproc. I see a `csl` field used in the skeleton, but it should work only when `citatation_package: "default"` (which will be pandoc's) and not with natbib. 

Should `natbib` been enforced like in `informs_article()` ? 

Pandoc citeproc have some benefit so it would be good if this is supported too.

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+\stitle{$it.title$}
+\sdescription{$it.description$}
+\end{supplement}
+$endfor$$endif$
+
+$if(acknowledgements)$
+\begin{acks}[Acknowledgments]
+$acknowledgements$
+\end{acks}
+$endif$
```
cderv, line 157: I don't know how convenient it would be, but as this content comes right after the body part with nothing in between, there is a solution where this could be removed from the template, and included directly in markdown by taking advantage of the Custom Blocks syntax. (https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html)

````markdown

::: {.supplement latex = 1}

\stitle{title}
\sdescription{
Some content
}

:::

::: {.acks, latex = "[Acknowledgement]"}
Some content
:::
````

For supplement, this is not ideal because of the special command for header and title, but this could be possible to use Lua filter to transform this markdown to the correct LaTeX

````markdown
::: {.supplement latex = 1}

# Title of Supplement A

Short description of Supplement A.

# Title of Supplement B

Short description of Supplement B.
:::
````

Would that be worth it ?

File: inst/rmarkdown/templates/isba/skeleton/LICENSE [hunk 8]
```diff
@@ -0,0 +1,415 @@
+The LaTeX Project Public License
```
cderv, line 1: Can we put this LICENCE file in a parent directory ? 

Does it really need to be copied to the user directory when creating a template  ?

inst/rmarkdown/templates/isba/skeleton/README.md
```diff
@@ -0,0 +1,36 @@
+## About
```
cderv, line 1: Same for this file - does it really need to be copied into the user directory ? 

I would put it one level above, or in the resource folder

inst/rmarkdown/templates/isba/skeleton/badoc.tex
```diff
@@ -0,0 +1,455 @@
+% \documentclass[ba,preprint]{imsart}% use this for supplement article
```
cderv, line 1: I don't think this file should be included in the **rticles** package. 

README says : 
> -   `badoc.tex` - LaTeX source file of the sample paper. 
    It contains useful information about typseting an article.

We don't need that, do we ? 

At least, we should add in Rbuildignore if we just want to keep a copy in Git.

inst/rmarkdown/templates/isba/skeleton/fig001.eps
```diff
@@ -0,0 +1,524 @@
+%!PS-Adobe-2.0 EPSF-2.0
```
cderv, line 1: I don't think we need this file. It is not use in the Rmd template. It is used in `badoc.tex` and we don't need that file I think. This should be remove IMO.

Same for the associated pdf file.

inst/rmarkdown/templates/isba/skeleton/sample.bib
```diff
@@ -0,0 +1,2244 @@
+%% BiBTeX database for models.tex bibliography
```
cderv, line 1: Do we use all references ? Is it possible to reduce the sample bib file to the bare minimum. Aim is to save space.

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,116 @@
...
+abstract: |
+  The abstract should summarize the contents of the paper. It should be clear, descriptive, self-explanatory and not longer than 200 words. It should also be suitable for publication in abstracting services. Please avoid using math formulas as much as possible.
+
+  This is a sample input file.  Comparing it with the output it generates can show you how to produce a simple document of your. That was the second paragraph.
+date: "`r Sys.Date()`"
+bibliography: sample.bib
+#linenumbers: true
+#numbersections: true
+csl: imsart.csl
+output:
```
cderv, line 75: We could also add the other format in the template 
````
output:
  rticles::isba_article: default
```` 

and not just add in the document that this is possible to use without **bookdown**

File: inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd [hunk 14]
```diff
@@ -0,0 +1,116 @@
...
+  Also thankful to God, mom and my cats.
+abstract: |
+  The abstract should summarize the contents of the paper. It should be clear, descriptive, self-explanatory and not longer than 200 words. It should also be suitable for publication in abstracting services. Please avoid using math formulas as much as possible.
+
+  This is a sample input file.  Comparing it with the output it generates can show you how to produce a simple document of your. That was the second paragraph.
+date: "`r Sys.Date()`"
+bibliography: sample.bib
+#linenumbers: true
+#numbersections: true
+csl: imsart.csl
```
cderv, line 74: This is normally used with Pandoc's citeproc only. I don't think it has any effect here. Also `imsart.csl` is not provided in the package it seems, so it would error with something like file not found. It needs to be a url or provided.

R/article.R
```diff
@@ -472,3 +472,34 @@ wellcomeor_article <- function(..., number_sections = FALSE, keep_tex = TRUE, ci
...
+#' * `doi` DOI of the article
+#' * `arxiv` Arxiv id
+#' * `acknowledgements` acknowledgement text, limited to 250 words
+#' * `bibliography` BibTeX `.bib` file
+#' * `longtable` set to `true` to include the `longtable` package, used by default from `pandoc` to convert markdown to LaTeX code
+#' * `header-includes`: custom additions to the header, before the `\begin{document}` statement
+#' * `include-after`: for including additional LaTeX code before the `\end{document}` statement
+#' @export
+#' @rdname article
+isba_article <- function(..., highlight = NULL, citation_package = "natbib") {
```
cderv, line 499: I think at least `keep_tex = TRUE` should be added and passed to `pdf_document_format()` - the default or user value would be used.

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,116 @@
...
+output: 
+  bookdown::pdf_book: 
+    base_format: rticles::isba_article
+    md_extensions: -autolink_bare_uris #keep these off or \printead{} will fail on email addresses
+    keep_tex: true # you need it to submit to the journal
```
cderv, line 79: This should be put as default in the format function as argument

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,116 @@
...
+$$a+b\geq c$$
+where $a$ is a variable and $c$ is variable as well.
+
+Dashes come in three sizes: a hyphen, an intra-word dash, a medium dash for number ranges like 1--2, and a punctuation dash in place of a comma, semicolon, colon or parentheses ---like this.
+
+The following are some other citations to check that the bibliographic style file is working correctly: \citet{akaike}, \citet*{akivarsq}, \citet{dyke},
+\citet{greene}, \citet*{kstuart}, \citet{hilbetech}, \citet{hilbe}, 
+\citet{hilbeglm}, \citet{maddalacntrsq}, and \citet*{companion}.  
+
+# References {#references .unnumbered}
```
cderv, line 115: FWIW, I also believe this will be useful for Pandoc citeproc. `natbib` will ignore it and use its own title (Pandoc seems clever enough to remove it when natbib is used)

inst/rmarkdown/templates/isba/template.yaml
```diff
@@ -0,0 +1,4 @@
+name: Bayesian Analysis article
+description: >
+    LaTeX related material for the manuscript preparation.
+create_dir: true
```
cderv, line 4:
```suggestion
name: Bayesian Analysis article (ISBA)
description: >
       Template for publishing to The International Society for Bayesian Analysis (ISBA)
create_dir: true
```

cderv (changes requested):
Hi, 

Thanks for the contribution. I have put my comment below. Happy to discuss the changes.

dmi3kno:
I see that it is failing on earlier versions of pandoc, likely because of the pandoc pipe `/pairs` I am using. I am happy to put a minimum version requirement on pandoc for this template. 
```r
pandoc_available("2.10.0", error=TRUE)
}
```
Is there a preferred way to do it?

cderv:
You can do this like in `informs_article()` that I shared in the comment https://github.com/rstudio/rticles/pull/461#discussion_r796487278

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
```diff
@@ -0,0 +1,116 @@
...
+  This is a sample input file.  Comparing it with the output it generates can show you how to produce a simple document of your. That was the second paragraph.
+date: "`r Sys.Date()`"
+bibliography: sample.bib
+#linenumbers: true
+#numbersections: true
+csl: imsart.csl
+output: 
+  bookdown::pdf_book: 
+    base_format: rticles::isba_article
+    md_extensions: -autolink_bare_uris #keep these off or \printead{} will fail on email addresses
```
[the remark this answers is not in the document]
dmi3kno: I felt that forcing `md_extensions` to be hardcoded in `isba_article()` would be too rigid and intransparent to the user. When it is provided in the . Rmd template, the user is aware of what's going on and can add more extensions, if needed. I think my comment explains it why it is needed.

inst/rmarkdown/templates/isba/resources/template.tex
```diff
@@ -0,0 +1,163 @@
...
+\newcommand{\CSLIndent}[1]{\hspace{\cslhangindent}#1}
+$endif$
+
+% Pandoc header
+$for(header-includes)$
+$header-includes$
+$endfor$
+
+$preamble$
+
```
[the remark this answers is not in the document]
dmi3kno: Sure. Removed.

inst/rmarkdown/templates/isba/resources/template.tex
dmi3kno: Sure. Runtitle is something that appears in the template. I envisage that when title is long a shorter version may need to be provided. I will add this to documentation of the `isba_article()`

inst/rmarkdown/templates/isba/resources/template.tex
dmi3kno: Done thank you.

inst/rmarkdown/templates/isba/resources/template.tex
dmi3kno: Sure.

inst/rmarkdown/templates/isba/resources/template.tex
dmi3kno: Do you suggest I remove it and rely on the built-in mechanism?

inst/rmarkdown/templates/isba/resources/template.tex
dmi3kno: Do you suggest I remove it and rely on the built-in mechanism?

inst/rmarkdown/templates/isba/resources/template.tex
dmi3kno: I clearly don't know enough to have a say about alternative referencing systems. Natbib is something that is enforced in the template. I am happy to enforce it as `informs_article()` does, but I am open to trying alternative suggestions.

inst/rmarkdown/templates/isba/skeleton/LICENSE
dmi3kno: No I will remove the file.

inst/rmarkdown/templates/isba/skeleton/badoc.tex
dmi3kno: Removed

inst/rmarkdown/templates/isba/skeleton/sample.bib
dmi3kno: I will slim this to 3 records.

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
dmi3kno: I feel that the default template is seriously lacking in internal referencing (figures and tables). I don't see a reason to write scientific paper without a bookdown. I am happy to be otherwise, of course, but every time i switched to default I struggled with something related to getting equations numbered and referenced, figures properly captioned and such.

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
dmi3kno: `imsart.csl` is in the skeleton directory.

R/article.R
dmi3kno: I did in the skeleton

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
dmi3kno: Again, I think when done this way it is visible. But I am happy to put it as default and document in the function as you suggest

inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd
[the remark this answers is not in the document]
cderv: Setting a default value for an argument seems to me something that would be transparent enough. I feel from your comment that this needs to be set otherwise it is not working. So I thought having the format setting it by default would be ok. 

````r
isba_article <- function(..., highlight = NULL, citation_package = "natbib", md_extensions = c("-autolink_bare_uris")) {
  pdf_document_format(
    "isba", highlight = highlight, citation_package = citation_package, md_extensions = md_extensions, ...
  )
}
````

But that will have the same effect as setting it explicitly in the YAML header. Just that there is a default. 

I let you decide.

File: inst/rmarkdown/templates/isba/resources/template.tex [hunk 21]
```diff
@@ -0,0 +1,163 @@
+% based on https://vtex-soft.github.io/texsupport.isba-ba/
+% \documentclass[ba,preprint]{imsart}% use this for supplement article
+\documentclass[ba]{imsart}
```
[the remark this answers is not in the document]
cderv: I see in the comment 
````
% \documentclass[ba,preprint]{imsart}% use this for supplement article
````

So I understood that `preprint` is an option that can be passed. 
Without the `classoption` variable it can't. 

If I misunderstood, please feel free to ignore. 

And in general, not all my remarks require change if you don't think that worth it. You know better this format and you are the user.

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 1 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/resources/template.tex:96]: > I envisage that when title is long a shorter version may need to be provided

Maybe we should call it `shorttitle` then to try reuse variable name that seems more suited from #411

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 4 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/resources/template.tex:144]: This need to be moved at the end of the document IMO so that it has the same effect as other template, and is coherent with the Pandoc default template. 

Pandoc's doc says about that 
> Include contents of FILE, verbatim, at the end of the document body (before the </body> tag in HTML, **or the \end{document} command in LaTeX**).

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 5 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/resources/template.tex:137]: Same as above, 

Pandoc doc says about this 
> Include contents of FILE, verbatim, at the beginning of the document body (e.g. after the <body> tag in HTML, **or the \begin{document} command in LaTeX**). 

So we should place it a bit before in the document, or just remove and this template won't support `includes$for before_body`.

File: inst/rmarkdown/templates/isba/template.yaml (hunk 18 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/template.yaml:4]: Done.

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 6 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/resources/template.tex:160]: Having an alternative using Pandoc citeproc and a CSL  would mean : 

- There is a CSL file that can be used to have the same expected style 
- `$if(natbib)$` should be added in the template around the part that does not need to be included if `natbib` is not used
- `citation_package: "default"` in the format function would trigger Pandoc citeproc to create the references sections below your last header with id `#references`

I did not find a CSL for ISBA in https://www.zotero.org/styles but I may not know the correct name. `ba.bst` is probably renamed from another style, maybe AMS one. 

Anyway, I am fine doing like `informs_article()` if we don't have the right option for this journal. 

Do you know if a `isba.csl` or `ba.csl` is available ?

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 7 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/resources/template.tex:157]: Oh I understand. Acknowledgement and Supplement are required and you assume that if the template have YAML field it should be kept. I was not doing the same assumption. Everything for me in a template can be removed, including YAML.  If user remove the YAML field, it will have the same effect than not including this in the body. But I agree it is less likely one will remove the field than removing the content. 

I was suggesting an option using Markdown in body because this template can support it and writing document content inside a YAML field is not really the easiest. I guess it depends on the content the formatting it should have. 

I am fine with your option if you prefer, I was just suggesting a way (not so well known) to create such LaTeX environment from Markdown, which makes it easier sometimes.

File: inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd (hunk 13 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd:75]: Yes we agree - it should not be a **bookdown** feature only. This will change in the future. 

Let's live it like that. I just suggested to put in addition as this is done for other template, and this would show how-to, instead of writing about it in the help page. But this is ok like that. 

Thanks for your answer.

File: R/article.R (hunk 15 above)
Comment (cderv) [reply R/article.R:499]: Why not do it by default then ? 

I understand this is working the same, but setting some default value seems to me like a nice gesture when this is expected. One need to know that `keep_tex: TRUE` needs to be set. I know there is the skeleton but it seems to me this could be a default. Also because other formats do that. 

Do you have strong opinion on not setting this by default ? You set the citation package and highlight by default. Why not `keep_tex` ? 

Just trying to understand your thinking as a direct user

File: inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd (hunk 16 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd:79]: I understand, but this could be done in both place. I commented in https://github.com/rstudio/rticles/pull/461#discussion_r796672256
I am trying to understand (and not that important) why this matter for `keep_tex` and not `highlight` or `citation_packages` where default have been set, and field not mentioned in YAML header.

File: inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd (hunk 17 above)
Comment (cderv) [reply inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd:115]: I can suggest some modification but only if we have some CSL to use as discussed in https://github.com/rstudio/rticles/pull/461#discussion_r796656273

I did not find it so maybe this journal does not have a Citation Style Language file available. 

Forcing natbib is ok - this formats will target PDF only so using `natbib` is enough.

File: inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd (hunk 19 above)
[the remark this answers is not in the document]
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd:79]: If you don't mind, I would like to keep it in YAML

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 4 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/resources/template.tex:142]: Coll will do

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 5 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/resources/template.tex:135]: Understood. I placed it before frontmatter.

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 6 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/resources/template.tex:158]: No and cant find any. I will enforce `natbib` then.

File: inst/rmarkdown/templates/isba/resources/template.tex (hunk 7 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/resources/template.tex:155]: Understood and appreciated. Putting things in YAML is annoying, but in this case I err on the side of caution. BA is really particular about supplementary material

File: inst/rmarkdown/templates/isba/skeleton/README.md (hunk 9 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/skeleton/README.md:1]: Removed

File: inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd (hunk 14 above)
Comment (dmi3kno) [reply inst/rmarkdown/templates/isba/skeleton/skeleton.Rmd:75]: Yes, I overlooked the extension (need glasses). Yes, CSL is missing. I will remove it from the skeleton.

File: NEWS.md [hunk 22]
```diff
@@ -3,6 +3,7 @@
 ## NEW FEATURES
 
 - New `informs_article()` template for submissions to INFORMS journals (thanks, @robjhyndman, #460).
+- New `isba_article()` template for submissions to Bayesian Analysis journal (thanks, @dmi3kno, ).
```
cderv, line 6:
```suggestion
- New `isba_article()` template for submissions to Bayesian Analysis journal (thanks, @dmi3kno, #461).
```

tests/testit/test-formats.R
```diff
@@ -50,6 +50,7 @@ test_format("ieee")
 test_format("ims")
 test_format("ims", output_options = list(journal = "aap"))
 test_format("informs", skip = !rmarkdown::pandoc_available("2.10"))
+test_format("isba")
```
cderv, line 53:
```suggestion
test_format("isba", skip = !rmarkdown::pandoc_available("2.10"))
```

R/article.R
```diff
@@ -472,3 +472,41 @@ wellcomeor_article <- function(..., number_sections = FALSE, keep_tex = TRUE, ci
...
+#' * `bibliography` BibTeX `.bib` file
+#' * `longtable` set to `true` to include the `longtable` package, used by default from `pandoc` to convert markdown to LaTeX code
+#' * `header-includes`: custom additions to the header, before the `\begin{document}` statement
+#' * `include-after`: for including additional LaTeX code before the `\end{document}` statement
+#' @export
+#' @rdname article
+isba_article <- function(..., keep_tex=TRUE, highlight = NULL, citation_package = "natbib") {
+    if (citation_package != "natbib") {
+      stop("ISBA template only supports `natbib` for citation processing.")
+    }
```
cderv, line 503:
```suggestion
  if (citation_package != "natbib") {
    stop("ISBA template only supports `natbib` for citation processing.")
  }
```

R/article.R
```diff
@@ -472,3 +472,41 @@ wellcomeor_article <- function(..., number_sections = FALSE, keep_tex = TRUE, ci
...
+  # from https://github.com/rstudio/rmarkdown/issues/372
+  #md_extensions <- c("+ascii_identifiers", "+tex_math_single_backslash", "-autolink_bare_uris")
+  if (!rmarkdown::pandoc_available("2.10")) {
+    stop("isba_article requires a minimum of pandoc 2.10.")
```
cderv, line 507:
```suggestion
    stop("`isba_article()` requires a minimum of pandoc 2.10.")
```

R/article.R
```diff
@@ -472,3 +472,41 @@ wellcomeor_article <- function(..., number_sections = FALSE, keep_tex = TRUE, ci
...
+#' * `doi` DOI of the article
+#' * `arxiv` Arxiv id
+#' * `acknowledgements` acknowledgement text, limited to 250 words
+#' * `bibliography` BibTeX `.bib` file
+#' * `longtable` set to `true` to include the `longtable` package, used by default from `pandoc` to convert markdown to LaTeX code
+#' * `header-includes`: custom additions to the header, before the `\begin{document}` statement
+#' * `include-after`: for including additional LaTeX code before the `\end{document}` statement
+#' @export
+#' @rdname article
+isba_article <- function(..., keep_tex=TRUE, highlight = NULL, citation_package = "natbib") {
```
cderv, line 500:
```suggestion
isba_article <- function(..., keep_tex = TRUE, highlight = NULL, citation_package = "natbib") {
```

cderv (approved):
Thanks. I'll commit the few small tweaks I have seen and merge. Thank you !

dmi3kno:
Could you please quickly correct the readme file. In the table the name of the function is `isba_article()` not `informs_article()`

cderv:
It is done. Thanks for seeing this!

PR #462, opened 2022-01-03
Add IOP template

robjhyndman:

- [x ] This project uses a Contributor Licence Agreement (CLA) that you'll be asked to sign when opening a PR. This is required for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). We use a tool called CLA assistant for that.  

- [x ] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [ x] Document your function using **roxygen2**. Markdown syntax is supported. Refer to <https://roxygen2.r-lib.org/articles/rd-formatting.html> for formatting references.

- [ x] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname/resources/template.tex`.

- [x ] Add a skeleton article `inst/rmarkdown/templates/journalname/skeleton/skeleton.Rmd`.

- [x ] Add a description of the template `inst/rmarkdown/templates/journalname/template.yaml`.

- [ x] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. If you are using TinyTeX or TeX Live, you can verify if a package is available on CTAN via `tinytex::parse_packages(files = "FILENAME"")` (e.g., when `FILENAME` is `plain.bst`, it should return `"bibtex"`, which means this file is from a standard CTAN package). Please keep the number of new files absolutely minimal (e.g., do not include PDF output files), and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include too many items in it without using all of them).

- [ x] Update Rd and namespace (could be done by `devtools::document()`).

- [x ] Update NEWS.

- [ x] Update README with a link to the newly supported journal. Please add your Github username and the full name of the journal (follow other examples in the list). 

- [x ] Add a test to `tests/testit/test-formats.R` by adding a line `test_format("journalname")`. We try to keep them in alphabetical order.

- [ x] Add your name to the list of authors `Authors@R` in DESCRIPTION. You don't need to bump the package version in DESCRIPTION.

Thank you!


robjhyndman:
Any changes required to this?

inst/rmarkdown/templates/iop/resources/template.tex
```diff
@@ -0,0 +1,122 @@
...
+\vspace{0.5cm}\noindent\textit{Keywords}: $for(keywords)$$keywords$$sep$, $endfor$
+
+\maketitle
+
+$if(twocolumn)$\ioptwocol$endif$
+
+$for(include-before)$
+$include-before$
+
+$endfor$
```
cderv, line 102: To be useful, this should be placed before  in the template. When just above `$body$` there is no difference between including a `include-before` or writing in the Rmd document first. 

From Pandoc's doc: 
> Include contents of FILE, verbatim, at the beginning of the document body (e.g. after the <body> tag in HTML, **or the \begin{document} command in LaTeX**).

So I think this should be place either before the `\title` or between Authors and abstract. Even if this is not to be used. Otherwise, we could not include it.

R/article.R
```diff
@@ -256,6 +256,16 @@ informs_article <- function(..., keep_tex = TRUE, citation_package = "natbib") {
   )
 }
 
+#' @section `iop_article`: Format for creating submissions to
+#'   IOP journals. Adapted from \samp{https://publishingsupport.iopscience.iop.org/questions/latex-template/}.
```
cderv, line 260: The guidelines provided in the link have lots of information in  `IOPLaTeXGuidelines.tex` example (and its pdf version) following your link. Should we precise in the help page here to look into the Guidelines there before submission ? 
Or the content could be included into the skeleton but this would duplicated. 

Just trying to simplify the step for users looking to use **rticles** for submissions and that won't look at the website expecting our template to be the only thing they need.

R/article.R
```diff
@@ -256,6 +256,16 @@ informs_article <- function(..., keep_tex = TRUE, citation_package = "natbib") {
...
+#' @export
+#' @rdname article
+iop_article <- function(..., keep_tex = TRUE, citation_package = "natbib") {
```
cderv, line 263: See my main remark on `citation_package`. Other value can be passed here, but I don't think the template will work in the current state.

inst/rmarkdown/templates/iop/resources/template.tex
```diff
@@ -0,0 +1,122 @@
...
+%% Adapted for rticles by Rob J Hyndman Rob.Hyndman@monash.edu. Dec 2021
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\documentclass[$if(fontsize)$$fontsize$$else$$if(twocolumn)$10pt$else$12pt$endif$,$endif$$if(papersize)$$papersize$,$endif$$for(classoption)$$classoption$$sep$,$endfor$]{iopart}
+
+%% BEGIN MY ADDITIONS %%
+\pdfminorversion=4
+\usepackage{hyperref, graphicx, booktabs}
+% Bibliography allowing natbib with Harvard
+\usepackage{har2nat}  % loads 'natbib' automatically
+\bibliographystyle{dcu}
```
cderv, line 12: This enforce the Harvard style in the template right ? 

Guidelines seems to also allow for a Vancouver style. Can this be changed ? 

> \item For the numerical (Vancouver) reference style we recommend that authors use 
 \verb"unsrt.bst"; this does not quite follow the style of published articles in our
 journals but this is not a problem.  Alternatively \verb"iopart-num.bst" created by Mark A Caprio
 produces a reference style that closely matches that in published articles.  The file is available from
\verb"http://ctan.org/tex-archive/biblio/bibtex/contrib/iopart-num/" .

Shouldn't this be configurable (even if a default is set ?) to use `unsrt.bst` or `iopart-num.bst` ? 

Could be with a default 

```
\bibliographystyle{$if(biblio-style)$$biblio-style$$else$dcu$$endif$}
```

Or more generic with settings in the YAML header: 
```
\bibliographystyle{$biblio-style$}
```
and
```yaml
# use dcu for alphabetical (Harvard) style references, 
# use unsrt or iopart-num for the numerical (Vancouver) reference style
biblio-style: dcu 
```

inst/rmarkdown/templates/iop/resources/template.tex
```diff
@@ -0,0 +1,122 @@
...
+
+$body$
+
+$if(acknowledgments)$
+\ack{$acknowledgments$}
+$endif$
+
+
+$if(bibliography)$
+\bibliography{$bibliography$}
```
cderv, line 112:
```suggestion
\bibliography{$for(bibliography)$$bibliography$$sep$,$endfor$}
```
in case there is several bibliography file 

From Pandoc's template: 
https://github.com/jgm/pandoc/blob/2fa8308afa1032b490018975d7138f0731369c3f/data/templates/default.latex#L512

cderv (changes requested):
Sorry for the delay on this one. I was focus on other packages the past month. I got bakc to **rticles** yesterday. 

Everything seems fine. I made small comments. 

Only thing I see if the reference mechanism: currently it is using `natbib`. Can this template be used with other system supported like  `biblatex` or Pandoc citeproc ? 
Compatibility would mean : 
* Adding the right variables (`natbib` and `biblatex` and if clause in the template
* Providing an example CSL file for Pandoc citeproc 
Most of the templates can use several usually. 

Thanks for the contribution ! And sorry again for the delay

R/article.R
robjhyndman: Added request to read guidelines carefully in the help file and in the template.

inst/rmarkdown/templates/iop/resources/template.tex
robjhyndman: Done

inst/rmarkdown/templates/iop/resources/template.tex
robjhyndman: Done

R/article.R
cderv: Thanks

cderv (approved):
Thanks a lot for taking care of all the changes. 

Last thing would be to CSL as a link to not include file within the R Package. We try to bundle as much as possible when the resource is available online and can directly be used. 
Having this in YAML would work fine: 

````
csl: https://www.zotero.org/styles/institute-of-physics-numeric
````

or

```
csl: https://www.zotero.org/styles/institute-of-physics-harvard
```

The part you wrote in the document Introduction could explain to download locally the CSL file if needed. 

Would that be ok with you to use URL for those file ? Thanks!

robjhyndman:
Sure. Done.

cderv (approved):

PR #465, opened 2022-02-07
Update Journal of Statistical Software template and class file

remlapmot:
I noticed that the JSS jss.cls file had been updated.

This PR

- updates jss.cls to use last year's updated version [here](https://www.jstatsoft.org/public/journals/1/jss-article-tex.zip)
- adds `\orcidlink{}` to each author if `orcid:` specified in the Rmd YAML header
- adds missing `\usepackage{}` line to template.tex from the JSS article.tex example file.

(Nb. I have no affiliation to the journal, I was just using the current **rticles** version and noticed that it was 1 version behind.)


cderv (approved):
Thank you !

PR #478, opened 2022-03-28
Copernicus template update to version 6.7

RLumSK:
Here is another Copernicus template update. @nuest 

- [X] Update NEWS.


cderv:
Thank you for maintaining this format up to date!

PR #479, opened 2022-03-29
Copernicus Template Update 6.8

RLumSK:
Sorry, and yet another update of the Copernicus template (@nuest). Yesterday, I must have just downloaded the latest version while an even newer version came probably online just minutes or hours later. 

Interesting: Now they finally support the LaTeX `listings` package, but adding "--listings" as an additional pandoc argument does not seem to work. Moreover, it is part of the *.cls file but not (yet) mentioned in their template file. We should watch it and add this with another update once also the LaTeX template mentions this option. 

- [X] Update NEWS.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## MINOR CHANGES
 
-- Update Copernicus Publications template to version 6.8 from 2022-03-28 (@RLumSK, #478).
+- Update Copernicus Publications template to version 6.8 from 2022-03-28 (@RLumSK, #478, #479).
 
 # rticles 0.23
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## MINOR CHANGES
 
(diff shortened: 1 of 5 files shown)


cderv:
>  adding "--listings" as an additional pandoc argument does not seem to work.

This is what is including in Pandoc's default template and maybe what needs to be included in this template for `--listing` to work.
https://github.com/jgm/pandoc/blob/7a7e1b2b70d7f22ad7ca2c6aea56055804a1cdb2/data/templates/default.latex#L223-L228

However, I don't know if you need to activate the Pandoc flag if the class already adds its own rules. Probably worth waiting the update of their template.  I'll merge this and you can do another PR when needed.

thanks !

RLumSK:
> However, I don't know if you need to activate the Pandoc flag if the class already adds its own rules. Probably worth waiting the update of their template. I'll merge this and you can do another PR when needed.

Thanks! I will do that.

PR #481, opened 2022-04-12
Fix elsevier issue with corresponding author

cderv:
Solves https://community.rstudio.com/t/corresponding-author-in-elsevier-template/134133

if `footnote` is not set for an author but `correspondingauthor` is, the `*` associated with the later should still be inserted


blongworth:
This addresses the issue I was seeing perfectly. Thank you!

cderv:
Then it is now merged. Thanks !

PR #487, opened 2022-05-23
Update Glossa article

stefanocoretta:
Fixing error in glossa article when rendering, caused by microtype.

The authors of the `glossa.cls` say in the comments that the microtype LaTeX package is used in production but it's fine if the authors don't have it. So the code loading microtype is now commented out. This solution is a workaround but works both for those using TinyTex and a non-TinyTex installation.


cderv (approved):
Thank you !

PR #503, opened 2022-08-22
Add a workflow to check new HTML5 checks on CRAN

cderv:
Just to help check as we got NOTES on CRAN


cderv:
It seems ok about the HTML5 check

PR #507, opened 2022-09-29
Enable natbib option for `asa_article()`

jepusto:
The documentation for `asa_article` indicates that it takes the `citation_package` option. However, the template did not provide any option to omit the natbib package.  This PR allows the natbib package to be included or omitted from the asa template, following the same format as the [`jasa_article` template](https://github.com/rstudio/rticles/blob/main/inst/rmarkdown/templates/jasa/resources/template.tex).


cderv:
Thanks ! 

I believe this is covered by #506 already. i'll check that this is the same change;

PR #519, opened 2023-03-17
Template update 7.2 -> 7.3

RLumSK:
And yet another Copernicus template update from version 7.2 to 7.3 @nuest 

- [X] Update NEWS.


cderv:
Thanks. I'll fix the GHA workflow independently

PR #557, opened 2024-01-03
Adapt template with Taylor & Francis Reference style S

cderv:
Taylor & Francis offers an other styles: http://www.tandf.co.uk/journals/authors/style/reference/tf_S.pdf and this goes with another template https://files.taylorandfrancis.com/InteractTFSLaTeX.zip

This is used by some journals like Journal of Applied Statistic https://www.tandfonline.com/action/authorSubmission?show=instructions&journalCode=cjas20#style-guidelines

This PR aims to adapt Taylor & Francis to this new style


cderv:
superseeded by #581

PR #568, opened 2024-05-25
Fix equal contribution in Springer template and add ORCID ID

nielsbock:
Thank you for a providing and maintaining a this great package!

While using the Springer Template, I noticed that the equal contributions line on the fronpage was missing. I have corrected it and tested it in my pull request so it should work now. Further, I have added an option to include ORCID ID at the end of each author line. Ideally, the ORCID ID should be placed after the star and dagger, but doing that is beyond my abilities. Supplying ORCID is not specified in the Springer Nature Template guide; however, some Springer Journals, e.g. the European Journal of Nutrition, accept manuscript submissions based on the Springer Nature Template but also encourages authors to provide ORCID. So I think it makes sense to add it.

I hope these changes are useful. I'm quite new in this field (this is only my second pull request ever) so I hope I haven't messed anything up.

EUR J NUTR submission guidelines:
https://link.springer.com/journal/394/submission-guidelines#Instructions%20for%20Authors_Title%20Page

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: rticles
 Title: Article Formats for R Markdown
-Version: 0.27.1
+Version: 0.27.2
 Authors@R: c(
     person("JJ", "Allaire", , "jj@posit.co", role = "aut"),
     person("Yihui", "Xie", , "xie@yihui.name", role = "aut",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # rticles (development version)
 
+- Fix `springer_article()` skeleton: It now uses `equal_contribution` in author field (thanks, @nielsbock, #568).
+
 # rticles 0.27
 
 - `joss_article()` now correctly works as `base_format` for `bookdown::pdf_book()` (thanks, @mlysy, #564).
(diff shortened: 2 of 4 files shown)


nielsbock:
I would also like to add that if numbered references is used for reference styles that allow it, the consecutive order references based on the position in the text is overuled by an alphabetical order. I have only found a manual solution which is to remove the captialized "SORT" that is occuring twice in the bst file (https://tex.stackexchange.com/questions/620603/problem-with-springer-bibliography-style-citation-style-numbered-sort-in-alph). Maybe there is a way to fix correct for this?

cderv:
Thanks ! 

> Further, I have added an option to include ORCID ID at the end of each author line. Ideally, the ORCID ID should be placed after the star and dagger, but doing that is beyond my abilities. Supplying ORCID is not specified in the Springer Nature Template guide;

Our template follows this Spinger Template: https://www.springernature.com/gp/authors/campaigns/latex-author-support
So tweaking for orcid will have impact on all further updates. but we'll make sure to not remove this. 

Too bad spinger does not handle it correctly. They have a `\orcid` command in their class file but the logo file associated is missing. 

Regarding position, their `\author` command does not seem to handle orcid, so I am not sure it can be put after the start and dagger. Spinger should be contacted to solve this.

PR #581, opened 2024-11-30
Add support to all bibliography styles required by Taylor & Francis journals

rlaboiss:
* R/article.R: Check whether the biblio-style field in the YAML header contains one of the allowed values
* inst/rmarkdown/templates/tf/skeleton/interact.cls: Update to version v1.05
* inst/rmarkdown/templates/tf/resources/template.tex: Choose the bibliography-related code according to biblio-style
* inst/rmarkdown/templates/tf/skeleton/tf{nlm,p,q,s}.bst: Add lacking bibliography style files from T&F
* man/article.Rd: Document the changes

- [ ] This project uses a Contributor Licence Agreement (CLA) that you'll be asked to sign when opening a PR. This is required for a significant pull request (it is fine not to sign it if a PR is only intended to fix a few typos). Unless you have done it in any other RStudio / Posit projects before, sign the individual or corporate contributor agreement as appropriate. You can send the signed copy to <contribute@rstudio.com>.

- [ ] Add the `journalname_article()` function to `R/article.R` if the output format is simple enough, otherwise create a separate `R/journalname_article.R`.

- [ ] Document your function using **roxygen2**. Markdown syntax is supported. Refer to <https://roxygen2.r-lib.org/articles/rd-formatting.html> for formatting references.

- [ ] Add the Pandoc LaTeX template `inst/rmarkdown/templates/journalname/resources/template.tex`.

- [ ] Add a skeleton article `inst/rmarkdown/templates/journalname/skeleton/skeleton.Rmd`.

- [ ] Add a description of the template `inst/rmarkdown/templates/journalname/template.yaml`.

- [ ] Please include the document class file (`*.cls`) if needed, but please do not include standard LaTeX packages (`*.sty`) that can be downloaded from CTAN. If you are using TinyTeX or TeX Live, you can verify if a package is available on CTAN via `tinytex::parse_packages(files = "FILENAME"")` (e.g., when `FILENAME` is `plain.bst`, it should return `"bibtex"`, which means this file is from a standard CTAN package). Please keep the number of new files absolutely minimal (e.g., do not include PDF output files), and also make examples minimal (e.g., if you need a `.bib` example, try to only leave one or two bibliography entries in it, and don't include too many items in it without using all of them).

- [ ] Update Rd and namespace (could be done by `devtools::document()`).

- [ ] Update NEWS.

- [ ] Update README with a link to the newly supported journal. Please add your Github username and the full name of the journal (follow other examples in the list). 

- [ ] Add a test to `tests/testit/test-formats.R` by adding a line `test_format("journalname")`. We try to keep them in alphabetical order.

- [ ] Add your name to the list of authors `Authors@R` in DESCRIPTION. You don't need to bump the package version in DESCRIPTION.

Thank you!


R/article.R
```diff
@@ -556,10 +556,69 @@ springer_article <- function(..., keep_tex = TRUE,  citation_package = "natbib",
...
+        "Invalid biblio_style in Taylor and Francis article. Allowed values are:",
+        paste(names(styles), collapse = ", ")
+      )
+    )
+  sty <- styles[[biblio_style]]
+  pandoc_args <- c(
+    pandoc_args,
+    rmarkdown::pandoc_variable_arg("bst-name", sty$bst),
+    rmarkdown::pandoc_variable_arg("biblio-commands", sty$cmd)
+  )
```
cderv, line 615: Really good idea to use this pandoc variable to fill the template accordingly ! 

a `switch(biblio_style, )` could also be use to make it clear this is a choice between a set of possibly value. though it does not probably improve much except readibility maybe

inst/rmarkdown/templates/tf/resources/template.tex
```diff
@@ -202,7 +197,7 @@
 $body$
 
 $if(bibliography)$
-\bibliographystyle{$if(biblio-style)$$biblio-style$$else$tfcad$endif$}
+\bibliographystyle{$bst-name$}
```
cderv, line 200: Why not use `biblio-style` here ? Why change to `bst-name` ? 

I think I would 

- Set `biblio-style` in the R function as you did.
- And add a check in the R function at the start to warn if `biblio-style` is already set in YAML metadata by user. Passing it as a variable will override the metadata. 
We do metadata check in pre_knit function for example 
https://github.com/rstudio/rticles/blob/fe7347dba3c4c54e262004ac943ce705c2c6efc2/R/ams_article.R#L45-L54 
or pre_processor
https://github.com/rstudio/rticles/blob/dfca82774dfd9bd9240a6217e2c4a11f582f6399/R/mdpi_article.R#L86-L93)

That would work right ?

inst/rmarkdown/templates/tf/resources/template.tex
```diff
@@ -1,15 +1,10 @@
...
 \documentclass[$for(classoption)$$classoption$$sep$,$endfor$]{interact}
 
 \usepackage{epstopdf}% To incorporate .eps illustrations using PDFLaTeX, etc.
 \usepackage{subfigure}% Support for small, `sub' figures and tables
 %\usepackage[nolists,tablesfirst]{endfloat}% To `separate' figures and tables from text if required
 
-\usepackage{natbib}% Citation support using natbib.sty
-\bibpunct[, ]{(}{)}{;}{a}{}{,}% Citation support using natbib.sty
-\renewcommand\bibfont{\fontsize{10}{12}\selectfont}% Bibliography support using natbib.sty
+$biblio-commands$
```
cderv, line 7: I really like this idea !! Thank you !

R/article.R
```diff
@@ -556,10 +556,69 @@ springer_article <- function(..., keep_tex = TRUE,  citation_package = "natbib",
...
+    stop(
+      paste(
+        "Invalid biblio_style in Taylor and Francis article. Allowed values are:",
+        paste(names(styles), collapse = ", ")
+      )
+    )
+  sty <- styles[[biblio_style]]
+  pandoc_args <- c(
+    pandoc_args,
+    rmarkdown::pandoc_variable_arg("bst-name", sty$bst),
```
cderv, line 613: I think we should use `biblio-style` here. See comment in template.tex

R/article.R
```diff
@@ -556,10 +556,69 @@ springer_article <- function(..., keep_tex = TRUE,  citation_package = "natbib",
...
 #' @rdname article
-tf_article <- function(..., keep_tex = TRUE, citation_package = "natbib") {
+tf_article <- function(..., keep_tex = TRUE, citation_package = "natbib",
+                       biblio_style = "CAD", pandoc_args = NULL) {
+  styles <- list(
+    APA = list(
+      bst = "apacite",
+      cmd = "\\usepackage[natbibapa]{apacite}
+\\setlength\\bibhang{12pt}
+\\renewcommand\\bibliographytypesize{\\fontsize{10}{12}\\selectfont}"
```
cderv, line 566: Would that be easier maintenance to store a character vector of line and then join them with new line in `pandoc_variable_arg()` call later ? 

I don't want this to break in the future because of accidental reformatting or else.

inst/rmarkdown/templates/tf/skeleton/interact.cls
cderv, line 1: Just to confirm: This is the update cls file without any modification by you for **rticles** right ?

inst/rmarkdown/templates/tf/skeleton/interactsample.bib
cderv, line 1: Could we reduce the content of this demo bib file to the minimum ? 

The previous one was quite short and this is because we have some size constraint in R packages. I don't want to have unnecessary content even if small text file at the end. 

Thanks !

cderv (changes requested):
Thanks a lot for this PR. Really good to see how the template can be improved. 

For the `biblio-style` logic, you could have also be processing depending on the value of the `biblio-style` 

Let's take this example 
````
output: tf_article
biblio_style: tfcad.bst
````
This following usual syntax. 

As a `pre_processor` in the format, you could check for `biblio-style` value, and check that `xfun::sans_ext(metadata$biblio-style)` a known value. If not you would error, if this is you would then define the appropriate `pandoc_variable_arg("biblio-commands")` associated with this style. 

This is an idea to keep the same YAML metadata as other format, while adding a specific processing to it. 

I am using `pre_processor` here, because this function returns new args to be used with Pandoc. 

> guess that, in case my PR get merged, my name will appear in the DESCRIPTION file. Please, tell me which role should I indicated for myself.

Please add yourself in DESCRIPTION as other contributor. 

> Should I add my name and affiliation in the YAML header of inst/rmarkdown/templates/tf/skeleton/skeleton.Rmd?

The skeleton is an example so no need to add you to it, unless you want to be the example. 

> There is only an annoying point: all tf*.bst are installed by

I don't think that is a big issue. A user will have to remove the undesired one. Otherwise, we need to think about this dynamically and only copy at render time the right file from resources if not already there. It  could be tricky but this is the only way without modifying general draft logic. 

Otherwise maybe I need to do a `rticles::create_article()` that would wrap the default draft, but do some postprocessing too depending on some specific rules 🤷 

I also triggered the test and they are failing. This is because documentation needs to be updated in Roxygen command. By creating a new file for this function it will solve it I thhink. 

Thanks again.

File: inst/rmarkdown/templates/tf/resources/template.tex [hunk 6]
```diff
@@ -1,15 +1,10 @@
-% interactcadsample.tex
-% v1.04 - May 2023
-
```
cderv, line 3: .cls file was update to v1.05 

No update in template with this version ?

rlaboiss:
> For the `biblio-style` logic, you could have also be processing depending on the value of the `biblio-style`
> 
> Let's take this example
> 
> ```
> output: tf_article
> biblio_style: tfcad.bst
> ```
> 
> This following usual syntax.
> 
> As a `pre_processor` in the format, you could check for `biblio-style` value, and check that `xfun::sans_ext(metadata$biblio-style)` a known value. If not you would error, if this is you would then define the appropriate `pandoc_variable_arg("biblio-commands")` associated with this style.
> 
> This is an idea to keep the same YAML metadata as other format, while adding a specific processing to it.
> 
> I am using `pre_processor` here, because this function returns new args to be used with Pandoc.

I need a clarification here. Are you proposing that we should only have the biblio_style metadata variable (containing the name of the `.bst` file), like this:
```
output: rticles::tf_article
biblio_style: tfcad.bst
```
instead of the scheme that I implemented, like this:
```
output:
  rticles::tf_article:
    biblio_style: CAD
```
or are you proposing to keep both possibilities?

rlaboiss:
> For the `biblio-style` logic, you could have also be processing depending on the value of the `biblio-style`
> 
> Let's take this example
> 
> ```
> output: tf_article
> biblio_style: tfcad.bst
> ```
> 
> This following usual syntax.
> 
> As a `pre_processor` in the format, you could check for `biblio-style` value, and check that `xfun::sans_ext(metadata$biblio-style)` a known value. If not you would error, if this is you would then define the appropriate `pandoc_variable_arg("biblio-commands")` associated with this style.
> 
> This is an idea to keep the same YAML metadata as other format, while adding a specific processing to it.
> 
> I am using `pre_processor` here, because this function returns new args to be used with Pandoc.
> 

I need a clarification here. Are you proposing that we should only have the biblio_style metadata variable (containing the name of the .bst file), like this:
```
output: rticles::tf_article
biblio_style: tfcad.bst
```
instead of the scheme that I implemented, like this:
```
output:
  rticles::tf_article:
    biblio_style: CAD
```
or are you proposing to keep both possibilities?

cderv:
> Are you proposing that we should only have the biblio_style metadata variable (containing the name of the .bst file)

Yes I am suggesting we could use the existing `biblio-style` metadata. it is usually how biblio style is set for articles with Pandoc and R markdwn. 

It could be `biblio_style: tfcad.bst` or `biblio_style: tfcad`


Do you think this is not great user experience ?

rlaboiss:
> > Are you proposing that we should only have the biblio_style metadata variable (containing the name of the .bst file)
> 
> Yes I am suggesting we could use the existing `biblio-style` metadata. it is usually how biblio style is set for articles with Pandoc and R markdwn.
> 
> It could be `biblio_style: tfcad.bst` or `biblio_style: tfcad`
> 
> Do you think this is not great user experience ?

Oh, this would change completely my design!

The existing `biblio_style` metadata is currently used for specifying the `.bst` file that is set as argument of the`\bibliography{}` command in the `template.tex` file.

The `bilbio_style` argument of my new `tf_article()` function, which can have values `"APA"`, `"CAD"`, `"NLM"`, `"TFP"`, `"TFQ"`, and `"TFS"`, serves another purpose. It is used for determining **both** the name of the `.bst` file **and** the necessary LaTeX commands, which will substitute the Pandoc construct `$biblio-commands$`.

I did this on purpose, such that the user does not need to know which specific `.bst` file they are required to use, according to the journal they are submitting to.

BTW, this is the reason why I used `$bst-name$ in the template, instead of the usual `$biblio-style$`.

Now, if I follow your suggestion, I would have to set `$biblio-commands$` inside `tf_article()` according to the value of the `biblio_style` in the metadata. Is that possible to do? If yes, I can change the code, even though I really prefer my design, which seems cleaner and less error-prone, in my humble opinion.

If you accept that my design is maintained, I could add a note in the documentation of `tf_article()` telling the user that the use of the `biblio_style` metadata is deprecated.

We might also change the name of the argument from `biblio_style` to `reference_style`, in order to avoid confusion (BTW, “Reference style” is the terminology used by Taylor & Francis). In this case, the user would use, in the YAML header, something like this:
```
output:
  rticles::tf_article:
    reference_style: APA
```

What do you think?

inst/rmarkdown/templates/tf/skeleton/interact.cls
rlaboiss: Yes, it is the upstream file, verbatim.

inst/rmarkdown/templates/tf/skeleton/interactsample.bib
rlaboiss: I merged the six files `interact{apa,cad,nlm,tfp,tfq,tfs}sample.tex` provided by Taylor & Francis. Each one has a different set of references in the `.bib` and this is the reason the final `interactsample.bib` file.

I will change the example in order to make the final `.bib` file minimal.

rlaboiss:
> We might also change the name of the argument from `biblio_style` to `reference_style`, in order to avoid confusion (BTW, “Reference style” is the terminology used by Taylor & Francis). In this case, the user would use, in the YAML header, something like this:
> 
> ```
> output:
>   rticles::tf_article:
>     reference_style: APA
> ```

In order to make things more concrete, I pushed the commit bbf912344c9a235ad3cd09a577419e34e0aa2ed0 containing changes along the lines above.  Here is the commit's log message:

**Deprecate the biblio_style option and rename the reference_style option**

This commit is intended to eliminate an ambiguity linked to the use of biblio_style as the name of the tf_article function option. In fact, biblio_style is traditionally used in the rticles package to indicate the name of the .bst file that will be used as an argument to the \bibliographystyle command in the LaTeX template.

The new version of tf_article accepts an option that determines the Taylor & Francis reference style to be used. Until now, the name biblio_style was used as the name of this option. In this commit, this name is changed to reference_style. This not only aligns with the terminology used by Francis & Taylor (“Reference Style”), but also avoids any confusion on the part of the user.

The tf_article function has also been modified to include a pre_knit function, in which the presence of the biblio_style in the YAML header is checked. If present, a warning message is issued concerning the depreciation of the biblio_style field.

cderv:
> Now, if I follow your suggestion, I would have to set $biblio-commands$ inside tf_article() according to the value of the biblio_style in the metadata. Is that possible to do? If yes, I can change the code, even though I really prefer my design, which seems cleaner and less error-prone, in my humble opinion.

Yes this is possible. In post processor, you check the `metadata$biblio-style` name, and you act accordingly if a known one to add the associated biblio command. 

This is quite similar to your current design except: 
- the name of the bst file needs indeed to be known instead of just reference style name. If you say the reference style name are more known by the authors, then that is good to have it like that even if it add a processing layer.
- the bibliocommand insertion is done in post processor instead of the function. 

The only drawback I see for your current design is that no custom `biblio-style` can now be used with `tf_article()`. So no custom tweak for a possibly unsupported journal and someone needing to make adjustment. 

I don't think that is common, but hiding behind `reference_style` will lead to that. I am ok with your last change though so we'll see if there are feedback on this.

cderv (approved):

rlaboiss:
> The only drawback I see for your current design is that no custom `biblio-style` can now be used with `tf_article()`. So no custom tweak for a possibly unsupported journal and someone needing to make adjustment.
> 
> I don't think that is common, but hiding behind `reference_style` will lead to that. I am ok with your last change though so we'll see if there are feedback on this.

Yes, I think that this situation (need to use another `.bst` file) will not arise in the normal case when submiting a manuscript to a T&F journal. Besides that, I really prefer the `biblo_style` deprecation instead of having to cope with conflicting cases, like this:
```
biblio_style: tfcad
output:
  rticles::tf_article:
    reference_style: NLM
```
Such a setup is almost guaranteed to produce broken results.

At any rate, I think that the PR is now ready for merging. I did extensive tests and the `skeleton.Rmd` file seems to produce correct PDF files for all the six reference styles.

cderv:
>  Besides that, I really prefer the biblo_style deprecation instead of having to cope with conflicting cases, like this:

To be clear on my thinking, this would not happen as you wouldn't need `reference_style`

It is either 
````markdown
biblio_style: tfcad
output: rticles::tf_article
````

and this would trigger CAD style command addition. No reference_style needed

Or 
````markdown
output:
  rticles::tf_article:
    reference_style: CAD
````
without needing `biblio_style`

This 
````markdown
biblio_style: tfcad
output:
  rticles::tf_article:
    reference_style: NLM
````
would be handled by `biblio_style` being ignored with `reference_style` taking precedence. 

> At any rate, I think that the PR is now ready for merging. I did extensive tests and the skeleton.Rmd file seems to produce correct PDF files for all the six reference styles.

This is awesome ! Thanks a lot for the work on this - greatly appreciated!

PR #586, opened 2025-01-29
Fixing a typo in  a comment in the Elsevier skeleton

setgree:
Right now it says
```
classoption: preprint, 3p, authoryear # remove authoryear is not using `elsarticle-harv`
```

I think it should be 
```
classoption: preprint, 3p, authoryear # remove authoryear if not using `elsarticle-harv`
```
is -> if


cderv:
Thank you

PR #588, opened 2025-03-10
Copernicus template update 7 10

RLumSK:
Yet another Copernicus template update, this time to version 7.10. Plus (inform @nuest):

- The folder contained a `.sty` file. I've removed that because it was a standard file (I've also tested it), and the pull request template asked not to include those files. 
- The pre-defined subsection headers, such as 'Disclaimer’, had disappeared in the PDF. The reason was that the macro in the `copernicus.cls` required a language flag (`english`, `german`, or `french`). Why this was not an issue before I cannot tell, but I think it is related to the babel package (without having further investigated the underlying change). I had also tried the LaTeX template directly available from Copernicus without using `rticles`, and I could reproduce the behaviour consistently. I have informed Copernicus via e-mail. Besides, the option to pick one of the three languages was always in the `copernicus.cls` but not used by the template. I have therefore added a language option to the skeleton template.

- [X] Update NEWS.


cderv (approved):
Thanks a lot. And for removing file. 

I am currently blocked to publish on CRAN a new release because the package is now more than 5MB built. And this is the limit, so removing some content is the next step.

PR #599, opened 2026-04-29
Update to JSS 3.6

cderv:
This is an update to JSS 3.6 to have a fix for an issue with recent TeX Live 

In relation with https://github.com/quarto-journals/jss/issues/13#issuecomment-4336717570


cderv:
JSS tests are passing

PR #600, opened 2026-04-29
Update gha

cderv:
To newest version of workflows


cderv:
Merging as I need to fix issues in format due to new change in package

